PR #18, opened 2013-04-23
Add support for parameter passed by reference

aitboudad:


TomasVotruba:
@aitboudad Thanks a lot for this fix. It helped me a lot!
I had to add it to my custom `Param` type, but thanks to your solution it was much faster :clap:

PR #26, opened 2013-09-04
fixes passed by reference issue

siad007:


TomasVotruba:
This should be fixed, as it happens and now is ignored.

#18 This is helpful and working solution.

PR #42, opened 2015-01-19
Add support for "Class[]" in method return type

vaidasm:
Adds support for type-hinted array in method return type


src/phpDocumentor/Reflection/DocBlock/Tag/MethodTag.php
```diff
@@ -76,7 +76,7 @@ public function setContent($content)
                 )?
                 # Return type
                 (?:
-                    ([\w\|_\\\\]+)
+                    ([\w\|_\\\\]+(?:\[\])?)
```
cebe: this does not allow a return type like `MyClass[]|int[]`.

mvriel:
Will be resolved using the new TypeResolver

luchaninov:
still not working - #50

Issue #27, opened 2013-12-01
Replace MarkdownExtra with Parsedown


mvriel:
See https://github.com/phpDocumentor/phpDocumentor2/issues/1080

mvriel:
All markdown references are removed in develop for v3

cebe:
:+1:

PR #45, opened 2015-02-23
Add support for the league/commonmark Markdown parser

colinodell:


cebe:
Note that markdown parsing should not be handled at this level in general, see https://github.com/phpDocumentor/ReflectionDocBlock/issues/27. Not sure if more parsers should be added here then...

mvriel:
I agree with @cebe that you should not rely on markdown parsing at this stage; it is only provided for backwards compatibility. phpDocumentor for example doesn't even use it anymore.

Nonetheless, given the precedent I will merge it now and attach a deprecated notice later because I think that is fair

colinodell:
Thanks, that sounds good to me.

composer.json
```diff
@@ -16,7 +16,8 @@
     },
     "suggest": {
         "dflydev/markdown": "~1.0",
-        "erusev/parsedown": "~1.0"
+        "erusev/parsedown": "~1.0",
+        "league/commonmark": "*"
```
GrahamCampbell: Whooa. What? Unbound version constraints are normally a really, really bad idea.

colinodell: I completely agree with your point, but I think it'll be okay in this specific case since the `CommonMarkConverter` class should **never** ever change, even between major versions.

That being said, I wouldn't be opposed to changing it.

mvriel: Also: please remember this is the 'suggest' section; the value after the package name is a free form text field and not a version number

GrahamCampbell: Oh yeh, didn't notice that. :)

PR #46, opened 2015-02-23
Remove composer.lock

colinodell:
Per the [Composer documentation](https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file):

> Note: For libraries it is not necessarily recommended to commit the lock file (...)

This is because the lock file will only affect people developing this library (not the end-user).  If `composer.json` allows consumers to use a newer version of the dependencies then we should also be testing against that.


mvriel:
In the past I would have advocated against this practice but given renewed insights and usage patterns I agree that the time has come to remove the lock file. Thank you for your contribution

PR #53, opened 2015-06-06
Make Context leaner and enable third parties to create them

mvriel:
Contexts are necessary for factories to resolve QSEN into FQSENs based
on partial namespaces and namespace aliases. These provide DocBlocks
with the namespace name and namespace aliases.

The new ContextFactory will enable third parties who don't use
phpDocumentor's Reflection component to construct a Context based on
a class reflector or namespace name (and file contents).


mvriel:
@jaapio can you review this item?

src/DocBlock/Context.php
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * Initializes the new context and normalizes all passed namespaces to be in Qualified Namespace Name (QNN)
+     * format (without a preceding `\`).
+     *
+     * @param string $namespace The namespace where this DocBlock resides in.
+     * @param array $namespaceAliases List of namespace aliases => Fully Qualified Namespace.
+     */
+    public function __construct($namespace, array $namespaceAliases = [])
+    {
+        $this->namespace = ('global' !== $namespace && 'default' !== $namespace)
```
jaapio: global and default should be class consts, But I don't get it yet why global and default are not allowed as namespace?

mvriel: Backwards compatibility with parts of phpDocumentor that have been there for a long time; I cannot predict what will break if I remove them. Perhaps after we have cleaned up the rest of the app?

src/DocBlock/ContextFactory.php
```diff
@@ -0,0 +1,174 @@
...
+     *
+     * @see Context for more information on Contexts.
+     *
+     * @return Context
+     */
+    public function createFromClassReflector(\ReflectionClass $class)
+    {
+        return $this->createForNamespace(
+            $class->getNamespaceName(),
+            file_get_contents($class->getFileName())
```
jaapio: This might conflict with applictions fetching files which are not directly accessable. An adapter could be a solution?

src/DocBlock/ContextFactory.php
```diff
@@ -0,0 +1,174 @@
...
+     */
+    public function createForNamespace($namespace, $fileContents)
+    {
+        $namespace = trim($namespace, '\\');
+        $useStatements = [];
+        $currentNamespace = '';
+        $tokens = new \ArrayIterator(token_get_all($fileContents));
+
+        while ($tokens->valid()) {
+            switch ($tokens->current()[0]) {
```
jaapio: is [0] always available?

tests/phpDocumentor/Reflection/DocBlock/ContextFactoryTest.php
```diff
@@ -0,0 +1,94 @@
...
+ * @license   http://www.opensource.org/licenses/mit-license.php MIT
+ * @link      http://phpdoc.org
+ */
+
+namespace phpDocumentor\Reflection\DocBlock {
+
+// Added imports on purpose as mock for the unit tests, please do not remove.
+    use Mockery as m;
+    use phpDocumentor\Reflection\DocBlock,
+        phpDocumentor\Reflection\DocBlock\Tag;
```
jaapio: please one namespace per use.

mvriel: Please read the comment above ;)

This is done on purpose because the test file is used as mock

src/DocBlock/ContextFactory.php
mvriel: [0] is, [1] and [2] not always

src/DocBlock/ContextFactory.php
mvriel: If you have a ReflectionClass instance than that means the file can be loaded using autoloading and thus it must be local; I do not immediately see how this situation can occur; can you think of a use-case?

mnapoli:
This looks good!

A little note: the code I used from Doctrine that parsed the class didn't read the whole file. I expect this to make a noticeable difference (reading whole file + `token_get_all()`). Here is the method that read only the part of the file that is above the class: https://github.com/PHP-DI/PhpDocReader/blob/master/src/PhpDocReader/PhpParser/UseStatementParser.php#L49-L67

I wonder though if their solution works with multi-namespace files.

mvriel:
@mnapoli I already found a bug in their solution that was not optimal :)

mnapoli:
ok great :)

mvriel:
@mnapoli also: i will be profiling this package before release; for phpDocumentor it is equally important that this is fast ;)

mnapoli:
Awesome!

barryvdh:
Why is the Context class moved from `src/phpDocumentor/Reflection/DocBlock/Context.php` to `src/DocBlock/Context.php`.  And the ContextFactory is also wrong (`src/DocBlock/ContextFactory.php`) This doesn't work with the PSR-0 autoloading you currently have configured.

Also, the tests are cleary failing:
> PHP Fatal error:  Class 'phpDocumentor\Reflection\DocBlock\ContextFactory' not found in ..

The same goes for my package using this library, class Context can't be found.. Could you please fix this issue?

src/DocBlock/Context.php
```diff
@@ -0,0 +1,75 @@
...
+    private $namespaceAliases = [];
+
+    /**
+     * Initializes the new context and normalizes all passed namespaces to be in Qualified Namespace Name (QNN)
+     * format (without a preceding `\`).
+     *
+     * @param string $namespace The namespace where this DocBlock resides in.
+     * @param array $namespaceAliases List of namespace aliases => Fully Qualified Namespace.
+     */
+    public function __construct($namespace, array $namespaceAliases = [])
```
barryvdh: Why is `$namespace` required? This is a BC break, the previous default was `''`

barryvdh: And the short array syntax isn't supported in 5.3. So if this was intentional, bump the composer.json requirements.

mvriel:
@barryvdh The changes to the composer.json are on my machine; I didn't want to submit a huge PR :)

Also: you might want to fix your package to a tag instead of dev-master; we are working towards version 2 of this component which will have significant BC breaks

mvriel:
In addition to the BC breaks might master have some instability in the next two weeks until the refactoring is complete

barryvdh:
@mvriel I have `2.0.x` as tag. Assuming this is SemVer, it should still work with `~2.0` even. But you have a branch-alias of `dev-master` to `2.0.x-dev`.

In my specific case, I do require a tag, but people are sometimes using a global dev-stability, eg. for testing Laravel 5.1 a few weeks before stable (with dev Symfony versions etc.)

If you plan to break BC, could you please update your branch-alias to 3.0.x, or use a seperate branch for the 3.0 changes?

mvriel:
@barryvdh version 2.0 has not been released so far; dev-master has the branch alias of `2.0.x-dev` because we are working towards version 2.0. If you have been working on the WIP 2.0 commits then you would probably need to fixate on a specific commit hash

mvriel:
I'm sorry that you are experiencing issues due to our work on the next version; is there an issue in 1.* that causes you not to use that?

barryvdh:
?? What do you mean? There a 5 published releases/tags on Github (https://github.com/phpDocumentor/ReflectionDocBlock/releases) and Packagist (https://packagist.org/packages/phpdocumentor/reflection-docblock), `2.0.0`, `2.0.1`, `2.0.2`, `2.0.3` and `2.0.4`

mvriel:
@barryvdh my apologies; I was mistaken with another component (I was deleting my comments because I just found out)

mvriel:
@barryvdh sorry about the mix up; I have changed the branch-alias of dev-master to 3.0 as you suggested. If all is well that should fix any issues that you experience?

barryvdh:
Yeah I guess so :) Thanks!

PR #54, opened 2015-06-07
Add support for types and improve resolution

mvriel:
As part of the re-architecting of phpDocumentor and its Reflection component
we needed to move the type classes from the \phpDocumentor\Descriptor
namespace to \phpDocumentor\Reflection\Types.

In addition to this I have removed the resolution / type creation from the
Collection class into a new Resolver class that is capable of resolving any
type into a Type or Fqsen object.


src/Type.php
```diff
@@ -0,0 +1,18 @@
...
+ * file that was distributed with this source code.
+ *
+ * @copyright 2010-2015 Mike van Riel<mike@phpdoc.org>
+ * @license   http://www.opensource.org/licenses/mit-license.php MIT
+ * @link      http://phpdoc.org
+ */
+
+namespace phpDocumentor\Reflection;
+
+interface Type
```
jaapio: missing dockblocks

src/Types/Array_.php
```diff
@@ -0,0 +1,63 @@
...
+ * @license   http://www.opensource.org/licenses/mit-license.php MIT
+ * @link      http://phpdoc.org
+ */
+
+
+namespace phpDocumentor\Reflection\Types;
+
+use phpDocumentor\Reflection\Type;
+
+class Array_ implements Type
```
jaapio: Should be final

src/Types/Resolver.php
```diff
@@ -0,0 +1,251 @@
...
+ * @link      http://phpdoc.org
+ */
+
+namespace phpDocumentor\Reflection\Types;
+
+use phpDocumentor\Reflection\DocBlock\Context;
+use phpDocumentor\Reflection\Fqsen;
+use phpDocumentor\Reflection\Type;
+
+class Resolver
```
jaapio: should be final

src/Types/Resolver.php
```diff
@@ -0,0 +1,251 @@
...
+{
+    /** @var string Definition of the ARRAY operator for types */
+    const OPERATOR_ARRAY = '[]';
+
+    /** @var string Definition of the NAMESPACE operator in PHP */
+    const OPERATOR_NAMESPACE = '\\';
+
+    /** @var string[] List of recognized keywords and unto which Value Object they map */
+    private $keywords = array(
+        'string' => 'phpDocumentor\Reflection\Types\String',
```
jaapio: it would be nice if you could add an addKeyword method, so this list can be extended by the application

jaapio:
In general it looks good. But the resolver is quite a massive class. I don't like the switch that you used to check which result should be returned. I would apply the same strategy pattern as we did for the descriptors in the reflection component. But that might be overkill? This reduces the number of methods in the resolver, and improves the readability a lot.

mvriel:
@jaapio it is ready for review; as discussed I have chosen not to split the Resolver into strategy objects because of two reasons:

1. Operator precedence, the order of things is quite important in this class as the wrong order will result in an incorrect result.
2. The class is not thus big and I do not expect it will be expanded by third parties, as such I think it will be overkill to separate the class into subobjects

Having said that: in the next PR I will extract the FQSEN resolving into an FQSEN factory because I need just that bit elsewhere without the Type resolution.

src/Type.php
jaapio: missing docblocks here.

src/Types/Array_.php
```diff
@@ -0,0 +1,87 @@
...
+     *
+     * @param Type $valueType
+     * @param Type $keyType
+     */
+    public function __construct(Type $valueType = null, Type $keyType = null)
+    {
+        if ($keyType === null) {
+            $keyType = new Mixed();
+        }
+        if ($valueType === null) {
```
jaapio: would prefer a white line between these lines

src/Types/Array_.php
```diff
@@ -0,0 +1,87 @@
...
+    /** @var Type */
+    private $keyType;
+
+    /**
+     * Initializes this representation of an array with the given Type or Fqsen.
+     *
+     * @param Type $valueType
+     * @param Type $keyType
+     */
+    public function __construct(Type $valueType = null, Type $keyType = null)
```
jaapio: I don't understand why $keytype is needed, basicly php only supports strings as keys?

src/Types/Resolver.php
```diff
@@ -0,0 +1,276 @@
...
+     * @param string $keyword
+     * @param string $typeClassName
+     *
+     * @return void
+     */
+    public function addKeyword($keyword, $typeClassName)
+    {
+        if (!class_exists($typeClassName)) {
+            throw new \InvalidArgumentException(
+                'The Value Object that needs to be created with a keyword "' . $keyword . '" must be an existing class'
```
jaapio: I prefer sprintf over string concats. But I think that is a personal choice.

jaapio:
A few small notes but overall it looks good for merge.

stof:
This PR should be partially reverted in favor of https://github.com/phpDocumentor/TypeResolver, otherwise class names will conflict (the revert should be done before releasing a new version of ReflectionDocBlock)

mvriel:
@stof I have just pushed most of my changes; which I know breaks the functionality of this package's develop branch. This has been deemed acceptable until v3 is released

PR #55, opened 2015-06-07
Move Context / ContextFactory to correct location

barryvdh:
Other location isn't autoloaded correctly. Fix for problems caused by PR #53


mvriel:
Moved to another component

PR #58, opened 2015-06-12
Bump php version

GrahamCampbell:
Would you consider bumping the version to 5.5.9+? This would make sense to me as it encourages people to use newer php versions and 5.5.9 is also the version of php that ubuntu 14.04 LTS ships with. Symfony 3.0 will be requiring 5.5.9+ and Laravel 5.1 released a few days ago requires 5.5.9 too rather than 5.5.0.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
         }
     ],
     "require": {
-        "php": ">=5.5",
+        "php": ">=5.5.9",
         "phpdocumentor/reflection-common": "~1.0@dev"
     },
     "autoload": {


cebe:
what is the reason to require 5.5.9 over 5.5.0?

GrahamCampbell:
Because I thought the entire point of dropping 5.4, aside from the new features of php, was to use a version of php which is still maintained, and 5.5.0 is very much not maintained, and has plenty of issues. I think it makes sense to force people up to 5.5.9.

cebe:
well, the main point of a version requirement is technical compatibility. If a library technically runs with 5.5.0 the requirement should be set this way. Whats the point in forbidding someone to use this lib on e.g. 5.5.6 if he is stuck on that version for some reason, if it technically works?
Upgrading PHP because of security issues is not the responsibility of a library version requirement but the system administator that runs the servers.

mvriel:
I agree with @cebe with the exception for web frameworks; when they know that a CVE is fixed in one of the later versions and know this impacts the user I can see that as an appropriate reason. 

This library does not suffer from these effects so unless there is a bug in one of the minor versions that prohibits this library from functioning I do not see a reason to prematurely raise the minimum version to 5.5.9

GrahamCampbell:
ok :)

PR #59, opened 2015-06-13
Feature/refactor to v2

mvriel:


.gitignore
```diff
@@ -1,3 +1,2 @@
 .idea
-composer.lock
```
GrahamCampbell: This file should be ignored in libraries.

mvriel: That advice is outdated, please see this post by Nils Adermann: https://blog.engineyard.com/2015/composer-continuous-integration

GrahamCampbell: I don't agree with that blog at all. The official advice from composer is still NOT to include it.

mvriel: I can understand your reluctance; I am currently partial to the information in the blog post since that provides information on which I can form an opinion. I am unaware of arguments against committing the lock file for a library.

Also, because I wanted to know, the [composer documentation](https://getcomposer.org/doc/02-libraries.md#lock-file) does not provide explicit advice on how to approach this; if anything they seem slightly partial to committing the lock file:

> For your library you may commit the composer.lock file if you want to. This can help your team to always test against the same dependency versions. However, this lock file will not have any effect on other projects that depend on it. It only has an effect on the main project.
>
> If you do not want to commit the lock file and you are using git, add it to the .gitignore.

What is your reasoning for not wanting a lock file in a library?

GrahamCampbell: 1. It makes maintaining a project with PRs and different branches hard because you'll get constant merge conflicts with the lock file.

2. Surely the point of tests is to check it **currently works** rather than worked with some old dependencies. You want to know if something broke straight away.

mvriel: > It makes maintaining a project with PRs and different branches hard because you'll get constant merge conflicts with the lock file.

I, personally, don't put much weight in the merge conflicts with the composer file with a library, this point is far more relevant in complete applications where the composer.json changes more regularly. If all is well a library's composer.json only changes seldomly.

> Surely the point of tests is to check it currently works rather than worked with some old dependencies. You want to know if something broke straight away.

This point is a bit contested; on one hand I agree with you but on the other hand should you, technically, be able to test with every version of a supported dependency.

To be frank: it can be considered poor for support to state in your composer.json that you accept a dependency between 1.0 and 2.0. Technically you'd have to test every released version to see if it doesn't break.

Once you accept that you can't support all versions of dependency you will need to take a point that you can support, as library author.

Here you have two choices: the latest or a version that you know was stable.

From a support perspective I prefer the latter, I don't want to know when a dependency breaks my build; I want to know when _I_ break the build.

mvriel: Lest I forget, thank you for providing your insights. I appreciate those as I always learn from such an exchange

cebe: > From a support perspective I prefer the latter, I don't want to know when a dependency breaks my build; I want to know when I break the build.

I have something to add to this point. If you test against the lock file which is ignored for other people you will not notice any break in some never version. This way the users of your library will notice and need to report it to you to fix it. If you always test against the latest version, you see breaking changes before they get reported to you and can react to them earlier.

Having the lock file means you test against a version combination that composer itself would never install (assuming there are newer versions out there already).

mvriel: That is indeed true but consider this: when do tests run? 

When a developer of contributor makes a change. 

So if the build breaks and you test without a composer.lock file you don't know whether it is a dependency that causes the break (which one?!) or the change you just did. While with a fixed composer.lock it is unambiguous that it is your change that broke the build and saves time debugging.

In an ideal situation you should monitor when dependencies update within your given version range and if that happens run the tests after having ran `composer update`, or by using `composer require --no-update`

cebe: I see your point, do you have a policy on when to run `composer update`? Should be done on a regular basis to catch both cases.

mvriel: Agreed; there is no policy currently but as we are typing I think we should set up a developer page and a few crons that check this for all repositories in the phpDocumentor organisation

DannyvdSluijs: Also take a look at versioneye. It can track outdated dependencies

cebe: :+1:

PR #68, opened 2015-12-22
Add mandatory description element to composer.json

willemstuursma:
composer validate gives the following warning: 

````
phpdocumentor/reflection-docblock is valid for simple usage with composer but has
strict errors that make it unable to be published as a package:
description : The property description is required
````


abacaphiliac:
+1

mvriel:
Thanks!

willemstuursma:
Thank you!

PR #72, opened 2016-02-26
Fix method name is empty when parenthesis is missing in @method

fabiang:
Fix method name is empty when parenthesis is missing in @method

### DIFF
--- a/src/DocBlock/Tags/Method.php
+++ b/src/DocBlock/Tags/Method.php
@@ -101,7 +101,9 @@ final class Method extends BaseTag implements Factory\StaticMethod
                 # Method name
                 ([\w\|_\\\\]+)
                 # Arguments
-                \(([^\)]*)\)
+                (?:
+                    \(([^\)]*)\)
+                )?
                 \s*
                 # Description
                 (.*)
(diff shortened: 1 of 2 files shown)


dunglas:
:+1:, I've tested it and this patch works for me.

dunglas:
ping @mvriel, any chance to get this one merged?

fabiang:
Merging this would help fixing some of my unit tests and ci builds, as this breaks mocking with Prophecy in some cases.

dunglas:
This bug prevent PHPUnit and PHPSpec to work with Symfony 3.1 (released today). If you merge the patch from @fabiang please create a new tag too.

GrahamCampbell:
@dunglas Is correct. This is a blocker for PHPUnit. This is affecting a lot of people.

Soullivaneuh:
@dunglas @GrahamCampbell Is this error I recently got on Travis related? https://travis-ci.org/sonata-project/SonataAdminBundle/jobs/133876692#L493

Only fail when Symfony 3.1 is installed and tested with PHPUnit 5.

Do you have a workaround waiting this issue?

stof:
@Soullivaneuh no it is not. Your issue is related to https://github.com/sonata-project/SonataAdminBundle/blob/daa65de93c11737481ab99475bb5b551234082ce/Route/RoutesCache.php#L62 not handling the case where ``$reflection->getFilename()`` does not give you a path, which happens in 2 cases:
- internal classes (will return false)
- eval'd code (will return a string containing ``eval()'d code`` among other things)

Symfony 3.1 is stricter about receiving a proper path in FileResource

Soullivaneuh:
@stof Ok sorry for the out of subject message and thank you for the tip. :+1:

scaytrase:
Any chance this gets merged?

ping @mvriel

dunglas:
https://twitter.com/mvriel/status/737584530370875393

JEDIBC:
:+1:

mvriel:
I have verified this fix and merged it; I am not entirely sure that this fix conforms spec (I find it hard to imagine what the exact impact is of removing the `()` since those are also a means to determine which part of the signature is the method name). But given the significant impact I am going to merge this anyway.

In a few minutes I will tag another release. 

Thank you for this contribution; it is good to see these incompatibilities addressed.

PR #73, opened 2016-04-07
Allow typed arrays in @method annotation.

hboomsma:
Since the type resolver has support for typed arrays,
I allowed the syntax for the @method annotation.

In effect you can now annotate magic methods like:
    @method Object[] myMethod($arg1, $arg2)
    @method int[] myOtherMethod()


mvriel:
Thank you for your contribution; I have merged it and will tag a new release in a few minutes

PR #75, opened 2016-06-08
Feature/php71 compat

hboomsma:
Renaming Void to Void_ since void is a reserved keyword in PHP7.1.

```
Running the unit tests with php nightly exposed the bug:

Fatal error: Cannot use phpDocumentor\Reflection\Types\Void as Void
because 'Void' is a special class name in
/home/travis/build/hboomsma/ReflectionDocBlock/tests/unit/DocBlock/Tags/MethodTest.php
on line 25
```

This PR is needed after phpDocumentor/TypeResolver#18 gets merged. I'll will update the composer.json
with the right version when needed. The build for this PR will obviously fail until then.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,13 @@ php:
   - 5.6
   - 7.0
   - hhvm
+  - nightly
 
 matrix:
   allow_failures:
-    - php: hhvm
+    - php:
+        - hhvm
+        - nightly
 
 cache:
   directories:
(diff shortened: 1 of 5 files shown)


GrahamCampbell:
:+1:

GrahamCampbell:
Can someone re-run travis please?

hboomsma:
I'll update the PR with the new release of  phpDocumentor/TypeResolver

mvriel:
@hboomsma can you update the composer.json as well to state `^0.2` for the TypeResolver dependency? This is because it won't work with older versions of TypeResolver and since this is a library people may try to include it

mvriel:
Looking good! Thank you for your contribution. I just merged it and will tag a build release in a moment

PR #82, opened 2016-08-23
Escape literal "%" because of vsprintf

### LINKED ISSUE #81: % in Description is not escaped
If there are placeholders in the description, there's a PHP warning issued, because the number of arguments doesn't match. This can be solved by escaping `%` with `%` in descriptions beforehand.

Example: https://github.com/symfony/console/blob/master/Question/ChoiceQuestion.php#L108

kelunik:
Fixes #81.


mvriel:
Thanks for the fix!

PR #84, opened 2016-09-28
Add support for rest argument notation

Seldaek:
`@method foo(...$arg)` breaks it as it tries to parse `...` as a type, and you end up with `InvalidArgumentException: "\...$arg" is not a valid Fqsen.`

This fixes it by ignoring `...` entirely. I suppose it would be nicer to keep track of it as an additional type or something, but it'd be nice if you can merge this as it fixes an issue already, and then work on improvements as needed.


mvriel:
Thanks for the help, Jordi. I agree that it would help to parse this and be able to show it in the docs but not breaking sounds like a good first step ;)

PR #88, opened 2016-11-14
Ignore URLs in @see tag, fixes #78

### LINKED ISSUE #78: URI not allowed in @see tag

Seldaek:
There are many more false-positives caused by Fqsen being greedily used in a lot of places, but this is an easy start that covers a lot of warnings out there.

### DIFF
--- a/src/DocBlock/Tags/See.php
+++ b/src/DocBlock/Tags/See.php
@@ -56,6 +56,10 @@ class See extends BaseTag implements Factory\StaticMethod
         $parts       = preg_split('/\s+/Su', $body, 2);
         $description = isset($parts[1]) ? $descriptionFactory->create($parts[1], $context) : null;
 
+        if (preg_match('{^https?://}', $parts[0])) {
+            return null;
+        }
+
         return new static($resolver->resolve($parts[0], $context), $description);
     }
 
--- a/examples/02-interpreting-tags.php
+++ b/examples/02-interpreting-tags.php
@@ -8,6 +8,7 @@ $docComment = <<<DOCCOMMENT
  * This is an example of a summary.
  *
  * @see \phpDocumentor\Reflection\DocBlock\StandardTagFactory
+ * @see https://example.org
  */
 DOCCOMMENT;
 


src/DocBlock/Tags/See.php
```diff
@@ -56,6 +56,10 @@ public static function create(
         $parts       = preg_split('/\s+/Su', $body, 2);
         $description = isset($parts[1]) ? $descriptionFactory->create($parts[1], $context) : null;
 
+        if (preg_match('{^https?://}', $parts[0])) {
+            return null;
```
cebe: it should still return something useful imo.

Seldaek: I'm not against it, but I have no clue what to return, I can't return a See instance as that expects a reference/Fqsen.

Seldaek: Another kinda related issue is `@var array<int, int>` also fails with `Invalid Fqsen array<int`, despite it being a valid type declaration.. I think Fqsen is just not flexible enough or something.

Seldaek:
Yup that definitely sounds superior. Would be good to hear from @mvriel 
though :)

src/DocBlock/Tags/See.php
felixfbecker: Only supporting `https?://` seems very limiting. What about people that omit the protocol?
Imo it should simply try to parse it as an FQSEN, and if it is not a valid FQSEN, return something like `Link` instance

Seldaek: @felixfbecker it's a good start that covers most cases I saw in the wild. I'm not against adding more but I can't promise I'll have time to do it.

staabm:
ping

SidIcarus:
Ping

glen-84:
@mvriel When will you have time to review this?

jaapio:
@felixfbecker We see @link as deprecated. People should use @see. I agree that Fqsen is not a valid solution here. I think the solution suggested by @cebe is a more suitable solution. I will try to implement this in a short time.

Thanks for the heads up!

sergeymakinen:
@jaapio take a look at #92 please. It seems that it's already done.

Webonaute:
it will be very nice to have a fix for it.

Simperfit:
ping

jaapio:
Thanks for this PR but I decided to implemented it in a other way. See #104

PR #89, opened 2016-11-14
Fix regex for tag name extraction and restore disabled test

Seldaek:
As far as I can tell, the goal of the regex is to extract `@<tag name>` followed by either whitespace and more stuff, or nothing at all (end of string). The new regex does that, and tests pass so I hope it's fine.


cebe:
looks good to me.

jaapio:
Thanks!

PR #90, opened 2016-12-22
fix: be able to parse @method tags with return types: type[] and $this

Fishdrowned:


jaapio:
Closing this in favor of #94

PR #92, opened 2016-12-27
Handle URIs in @see tags (fixes #78)

### LINKED ISSUE #78: URI not allowed in @see tag

sergeymakinen:
Fixes #78 / Replaces #88 

This PR adds an URI support to `@see` tags - the tag class will return `string|Fqsen`. That's it.
Tests are also here. :)


felixfbecker:
This is a breaking change.

jaapio:
Thanks for this PR but I decided to implemented it in a other way. See #104

PR #94, opened 2017-01-19
Allow `$this` as return type in `@method` annotations

### LINKED ISSUE #93: Return type $this in @method annotations not working
An annotation like

```php
/*
 * @method $this myMethod() My Description.
 */
```

leads to errors. I debugged and found out that the Regex in `phpDocumentor\Reflection\DocBlock\Tags\Method::create()` does not account for `$this` as return type, since it does not match the character `$` (only relevant part of regex):

```
[...]
# Return type
(?:
    (?:[\w\|_\\\\]+)
    # array notation           
    (?:\[\])*
)
[...]
```

The regex should be altered to allow `$this` as return type in `@method` annotations

Christoph-Harms:
Fixes #93 

- Modify the RegEx, so that an annotation like `@method $this myMethod()` is parsed correctly
- Write a Unittest for that usecase

### DIFF
--- a/src/DocBlock/Tags/Method.php
+++ b/src/DocBlock/Tags/Method.php
@@ -128,7 +128,7 @@ final class Method extends BaseTag implements Factory\StaticMethod
         $returnType  = $typeResolver->resolve($returnType, $context);
         $description = $descriptionFactory->create($description, $context);
 
-        if ('' !== $arguments) {
+        if (is_string($arguments) && strlen($arguments) > 0) {
             $arguments = explode(',', $arguments);
             foreach($arguments as &$argument) {
                 $argument = explode(' ', self::stripRestArg(trim($argument)), 2);
--- a/src/DocBlock/Tags/Method.php
+++ b/src/DocBlock/Tags/Method.php
@@ -91,10 +91,14 @@ final class Method extends BaseTag implements Factory\StaticMethod
                 )?
                 # Return type
                 (?:
-                    (
-                        (?:[\w\|_\\\\]+)
-                        # array notation           
-                        (?:\[\])*
+                    (   
+                        (?:[\w\|_\\\\]*\$this[\w\|_\\\\]*)
+                        |
+                        (?:
+                            (?:[\w\|_\\\\]+)
+                            # array notation           
+                            (?:\[\])*
+                        )
                     )?
                     \s+
                 )?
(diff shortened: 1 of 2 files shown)


Christoph-Harms:
Damnit, one unittest failing, but only with php 7.2.0 (nightly build). I'll figure it out.

Christoph-Harms:
OK, so the problem is located in `\phpDocumentor\Reflection\DocBlock\Tags\Method:131ff` (in the updated code. In the old code it's line 127ff):

```php
if ('' !== $arguments) {
    $arguments = explode(',', $arguments);
    foreach($arguments as &$argument) {
        $argument = explode(' ', self::stripRestArg(trim($argument)), 2);
        if ($argument[0][0] === '$') {

             [...]

        }
    }
}
```

If the `@method` annotation has no arguments, for *all* php versions *but the nightly build*, `'' !== $arguments` is false. Only with the php nightly build, it will somehow be true, leading to an error when evaluating `$argument[0][0] === '$'` four lines later.

I wanted to debug this, but there seems to be no way for me to get the php nightly build, so I can't.

I recommend merging this anyway, since there clearly is no wrong behaviour here.

Christoph-Harms:
Definitely gonna need your help here. Can't find out what's going wrong due to lack of php 7.2.

Christoph-Harms:
Alright, last change seems to have done the trick. I replaced

```php
if ('' !== $arguments)
```
with
```php
if (is_string($arguments) && strlen($arguments) > 0)
```
which basically means the same, with the only difference being that php 7.2 seems to be able to understand the latter, but not the first. 😕

FIA2008:
Is this already merged, if not, when can we expect this? The bug is pretty annoying to be honest.

jaapio:
`$this` is actually not a type. So I'm a bit unsure if I'm willing to accept this. Does this same issue occur when you use `self`?

jaapio:
I talked about this with @mvriel some time ago. Still think using `$this` as a type in docblocks is wrong. But since we are already supporting it we decided to add it back again.

Christoph-Harms:
So glad this got merged! <3

Although I get your point, `$this` is not a type. But since [the PHPDoc Standard](https://www.phpdoc.org/docs/latest/guides/types.html#keywords) explicitly allows it as a "type", I thought it would be inconsistent to not parse it correctly. Also, it is useful to differentiate between the two cases where a method returns another instance of the same class vs. the exact same object like in fluent interfaces.

PR #96, opened 2017-03-01
Custom tag formatter

yunosh:
This PR adds to the Serializer an option to use a custom tag formatter, other than the default PassthroughFormatter. It also adds an alternative formatter that aligns the tag values.


jaapio:
Can you please provide unittest for this change?

jaapio:
Thanks for this contribution!

PR #97, opened 2017-03-01
Fix author tag generation

yunosh:
When generating author tags, the author name and author email run into each other. If the email is empty, empty arrow brackets are generated. Both is fixed with this PR.


jaapio:
Could you please provide a testcase for this?

cebe:
The tests should also cover the case where author email is empty.

jaapio:
Thanks @yunosh

PR #98, opened 2017-04-11
Moved webmozart/assert to require-dev.

frankdejonge:
I saw this in my vendor folder, figured it wasn't needed.


frankdejonge:
Hmmm. Seems it is :(

PR #99, opened 2017-04-13
Add gitattributes file

jdufresne:
Used to allow composer to ignore tests and other development files in
distribution of releases.


jaapio:
Thanks!

PR #102, opened 2017-07-13
Unit tests fixed

akireikin:
## Motivation

Master is broken. 

## Changes

1. `AlignFormatterTest` class name fixed. Build is going to be green.
2. All necessary `@covers` and `@uses` tags added. No more risky tests.


jaapio:
Thanks!

Issue #74, opened 2016-05-14
Methods with empty body part are not ignored


ArSn:
In the following example:

``` php
/**
 * @method
 */
```

The parsing in `\phpDocumentor\Reflection\DocBlock\Tags\Method::create()` asserts at first that the method body shall not be empty (type & name), and hence, throws an exception, if it is empty.

Prior to 3.0, empty `@method` tags like the above were just ignored. Is there some reason why that is hard asserted to not be an empty string now?

If not, I'd love to create a pull request to change this (and make sure the method can properly handle an empty string as well). 

I am requesting this in order to be able to bump the dependency for phpspec/prophecy to ~3.0.

jayS-de:
There is also another issue connected to this one. When the magic method doesn't have a return type hint. For example:

```
/**
 * @method getValue()
 */
```

This also throws an InvalidArgumentException. See also phpspec/prophecy#277

ArSn:
Any info on this?

Gert-dev:
I just bumped into the issue mentioned by @jayS-de and got confused about whether this syntax was actually valid. Looking it up in the phpDocumentor documentation, the `Syntax` seems to indicate that the return type is required, but the `Description` then explicitly mentions that the return type is optional and `void` is then implied. [1]

[1] https://phpdoc.org/docs/latest/references/phpdoc/tags/method.html

jaapio:
I implemented a fix for the issue mentioned by @jayS-de. But I'm not sure if we want to support the example of @ArSn. Basically `@method` requires a method name. Only the `@method` tag has no meaning for us. 

@ArSn can you provide a use-case for `@method` where it doesn't require a name?

jaapio:
Closing this issue. due a leak of feedback from the creator. 

Feel free to reopen this issue when needed.

PR #103, opened 2017-07-14
Allow omitting method return type

jaapio:
Support for `@method myMethod()` was missing but according to the
docs of phpdocumentor this is possible.

Refs #74


jaapio:
thanks for the review @mvriel, think this can be merged?

PR #104, opened 2017-07-14
Allow see tag to url

### LINKED ISSUE #78: URI not allowed in @see tag

jaapio:
The see tag allows Fqsen and uri's. But in the current implementation
only Fqsen where allowed. This wrapps the uri and Fqsen in an uniform class
to be able address them in the same way.

Fixes #78
Replaces #88 and #92


jaapio:
@mbed67 I added you since you know if we are still supporting all features of the `@see` tag in here.

PR #107, opened 2017-07-18
Surround the creation of the tags with a try catch block

alexandre-le-borgne:
Surround the creation of the tags of a try catch block to support php doc with unknown tags such as @JMS\Type("int"), etc


Johnmeurt:
This is useful to fix the support of unkonwn tags but it is not the solution. I was thinking about those 2 solutions:
- Raise custom exception as UnknownTagException. This could be catch on the DocBlockFactory (same way you did in your commit).
- Adding an options management in the context. Then adding "ignore_unkonwn_tags" option. This allows the dev to choose the behavior.

jaapio:
I'm clossing this in favor of #110 which is an improved to support annotations in the future.

PR #110, opened 2017-08-04
Restores future support for annotations

### LINKED ISSUE #108: Release 3.2.0 fails when parsing annotations
With the release of 3.2.0 (more specific the changes from #89) parsing a docblock fails if it not contains only phpdoc but also annotations (for example, to configure Doctrine mapping metadata). 

For example, in Symfony we have the following test data:

```php
class Dummy extends ParentDummy
{

    /**
     * @var \DateTime[]
     * @Groups({"a", "b"})
     */
    public $collection;

    // ...
}
```

I am not sure what could be a possible fix here. Silently ignore these lines? But this may be bad user experience as it does not help to find your own mistakes. Maybe ignore only tags that are followed by parentheses?

jaapio:
In #89 the regex to parse tags was to strict. The first character after a tag
had to be a single string. Annotations typically start with a parentheses. Only character not
allowed after a tag is a [.

Fixes #108

### DIFF
--- a/src/DocBlock/StandardTagFactory.php
+++ b/src/DocBlock/StandardTagFactory.php
@@ -113,6 +113,12 @@ final class StandardTagFactory implements TagFactory
 
         list($tagName, $tagBody) = $this->extractTagParts($tagLine);
 
+        if ($tagBody[0] === '[') {
+            throw new \InvalidArgumentException(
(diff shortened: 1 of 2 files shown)


xabbuh:
Thanks, this fixed it for us.

Issue #111, opened 2017-08-05
Notice:  Uninitialized string offset: 0 in version 3.2.1


forgie1:
I'm getting cca 50% of times the same code is run here https://github.com/phpDocumentor/ReflectionDocBlock/commit/183824db76118b9dddffc7e522b91fa175f75119#commitcomment-23494690 Notice: `Uninitialized string offset: 0`.

`$this->tagFactory->create(trim($tagLine), $context);` was called with arguments:

```
$tagLine = "@ORM\Entity" (11)
$context = phpDocumentor\Reflection\Types\Context #01 
                    namespace private => "Plugins\Doctrine\Products" namespaceAliases private => array (2) 
                           ORM => "Doctrine\ORM\Mapping" 
                           Entities => "Etten\Doctrine\Entities"
```

in phpDocumentor\Reflection\DocBlockFactory:231 , resulting in the notice.

ahilles107:
I have this same error but in other place: 
```
notice: Uninitialized string offset: 0 in
      vendor/phpdocumentor/reflection-docblock/src/DocBlock/StandardTagFactory.php
      line 116
```

forgie1:
@ahilles107 yes, that is the same line I'm getting this error: https://github.com/phpDocumentor/ReflectionDocBlock/commit/183824db76118b9dddffc7e522b91fa175f75119#commitcomment-23494690

esynaps:
Same here, i'm adding "()" to all tags => it works

^@([\w\-\_\\\\]+)(?:\s*([^\s].*)|$)

@JMS\Exclude => 
Full match | 0-12 | `@JMS\Exclude`
Group 1. | 1-12 | `JMS\Exclude`

@JMS\Exclude()
Full match | 0-14 | `@JMS\Exclude()`
Group 1. | 1-12 | `JMS\Exclude`
Group 2. | 12-14 | `()`

temp:
This is definitely a BC break. 
This was introduced in StandardTagFactory.php, the check for $tagBody[0] being "[", but with no Annotations parameters (and no brackets) this value is an empty string, meaning that index 0 of the string is not set.
[Compare 3.2.0 to 3.2.1](https://github.com/phpDocumentor/ReflectionDocBlock/compare/3.2.0...3.2.1)
This can be easily fixed by checking for the value not being empty:
```php
if (!empty($tagBody[0]) && $tagBody[0] === '[') {
```

jaapio:
[...]

PR #112, opened 2017-08-07
Check tag body is an array before trying to access it as one

respinoza:
Related to issue #111 

It seems when the extract tag parts, it can return an empty string instead of an array (seems to be like that for years now). I added a simple check that it is in fact an array before accessing it.


respinoza:
Would it be possible to also have this fix in 3.x ?

respinoza:
I will create a new one as we should be checking for strlen

PR #113, opened 2017-08-07
Check tag body has a value before trying to access it as an string array

respinoza:
Related to issue #111 

Check if the string is not empty before accessing it as an array. After I noticed all the other failures in the tests from the previous MR, I understood what was the intention.


respinoza:
Again, would be nice to have this fix also in 3.2.x

ahilles107:
👍

remicollet:
+1

/cc @siwinski

jaapio:
I just released 3.2.2 and 4.0.1. Everybody should be ready to go. Please let me know when you have any issues regarding this issue.

Thanks for this contribution @respinoza!

PR #122, opened 2017-09-11
Update Description.php

Warxcell:
Getter for `tags`.

### DIFF
--- a/src/DocBlock/Description.php
+++ b/src/DocBlock/Description.php
@@ -69,6 +69,16 @@ class Description
         $this->bodyTemplate = $bodyTemplate;
         $this->tags = $tags;
     }
+    
+    /**
+     * Returns the tags for this DocBlock.
+     *
+     * @return Tag[]
+     */
+    public function getTags()
+    {
+        return $this->tags;
+    }
 
     /**
      * Renders this description as a string where the provided formatter will format the tags in the expected string
--- a/tests/unit/DocBlock/DescriptionTest.php
+++ b/tests/unit/DocBlock/DescriptionTest.php
@@ -65,6 +65,58 @@ class DescriptionTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * @covers ::getTags
+     * @uses \phpDocumentor\Reflection\DocBlock\Tags\Generic
+     * @uses \phpDocumentor\Reflection\DocBlock\Tags\BaseTag
+     */
+    public function testDescriptionTagsGetter()
+    {
+        $body = '@JoinTable(name="table", joinColumns=%1$s, inverseJoinColumns=%2$s)';
+
+        $tag1 = new Generic('JoinColumn', new Description('(name="column_id", referencedColumnName="id")'));
+        $tag2 = new Generic('JoinColumn', new Description('(name="column_id_2", referencedColumnName="id")'));
+
+        $tags = [
+            $tag1,
+            $tag2,
+        ];
+
+        $fixture = new Description($body, $tags);
+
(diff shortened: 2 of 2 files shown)


Warxcell:
There is `[getTags()](https://github.com/phpDocumentor/ReflectionDocBlock/blob/master/src/DocBlock.php#L157)` in DocBlock,but not here. Why?

jaapio:
Thanks again for your contribution. I see no harm to merge this. Not sure why you would need this. Since there is no way to get descriptor body template as well. Basically the template and the tags are the content of a description. So tags are for internal use only. 

Could you please provide a unit test for this function. It is way to easy but we want to have all lines covered. :-)

Warxcell:
@jaapio I need this because I'm converting Doctrine ORM Annotations from Simple to Non-simple. And I have this docblock:

```
 /**
     * @ManyToMany (targetEntity="Scope")
     * @JoinTable (name="oauth2server_authorization_codes_to_scopes",
     * joinColumns={@JoinColumn (name="auth_code_id", referencedColumnName="id")},
     * inverseJoinColumns={@JoinColumn (name="scope_id", referencedColumnName="id")}
     * )
     */
```

I'm using following code:

```
 private function convertToOrm(\phpDocumentor\Reflection\DocBlock\Tag $tag)
    {
        return 'Doctrine\ORM\Mapping\\'.$tag->getName();
    }

    private function isOrmTag(\phpDocumentor\Reflection\DocBlock\Tag $tag)
    {
        return class_exists($this->convertToOrm($tag));
    }

    public function visit(\phpDocumentor\Reflection\DocBlock $docBlock)
    {
        $newTags = $docBlock->getTags();

        foreach ($docBlock->getTags() as $i => $tag) {
            if ($tag instanceof \phpDocumentor\Reflection\DocBlock\Tags\Generic && $this->isOrmTag($tag)) {
                $description = $tag->getDescription();

//                if ($description) {
//
//                }

                $newTags[$i] = new \phpDocumentor\Reflection\DocBlock\Tags\Generic(
                    $this->convertToOrm($tag), $description
                );
            }
        }

        return new \phpDocumentor\Reflection\DocBlock(
            $docBlock->getSummary(),
            $docBlock->getDescription(),
            $newTags,
            $docBlock->getContext(),
            $docBlock->getLocation(),
            $docBlock->isTemplateStart(),
            $docBlock->isTemplateEnd()
        );
    }
```


After that docblock looks like that:

```
 /**
     * @Doctrine\ORM\Mapping\ManyToMany (targetEntity="Scope")
     * @Doctrine\ORM\Mapping\JoinTable (name="oauth2server_authorization_codes_to_scopes",
     * joinColumns={@JoinColumn (name="auth_code_id", referencedColumnName="id")},
     * inverseJoinColumns={@JoinColumn (name="scope_id", referencedColumnName="id")}
     * )
     */
```

I don't have way to replace `@JoinColumn` inside `@JoinTable` tag, because `@JoinColumn` is tag inside description of `@JoinTable`.

jaapio:
Seems plausible. However you could have done it by passing a custom formatter to the render method as well.

Warxcell:
@jaapio I was using version 3.1.1 and there is no way to change the formatter inside serializer. I will upgrade and try to fix it. 
BUT anyway - formatter should not alter data (I think) - It's not its responsibility.

Warxcell:
@jaapio I added two tests. 
I also cannot upgrade, because ~3.2.0 requires ` "phpdocumentor/reflection-common": "^1.0@dev"`
which conflicts with my minimum stability flag.

PR #126, opened 2017-11-03
Allow php >= 5.5

jurruh:
In commit 007a536bc5ce6d88a49c6655d0cf0a356337945f support for php 5.5 is dropped. I don't think this was the intention of that commit, this will fix it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "php": "^7.0",
+        "php": ">=5.5",
         "phpdocumentor/reflection-common": "^1.0.0",
         "phpdocumentor/type-resolver": "^0.4.0",
         "webmozart/assert": "^1.0"


jaapio:
Actually it was we do not support 5.5 anymore. Please upgrade to php 7

jurruh:
In tag 3.2.2 php 5.5 was still supported. According to http://semver.org/#spec-item-8, breaking backwards compatibility has to be a major update.

I understand that php 5.5 is deprecated, so maybe 5.6 is an option?

jurruh:
@jaapio did you read my previous comment? I would be happy to create a pull request for php 5.6 support.

jaapio:
Hi @jurruh,

I think we could return to add support for 5.6. I will accept a pr for that.

jurruh:
Great, I created one!

PR #127, opened 2017-11-10
Allow php >= 5.6

jurruh:
Allows support for php 5.6 or above

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "php": "^7.0",
+        "php": ">=5.6",
         "phpdocumentor/reflection-common": "^1.0.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "php": "^7.0",
+        "php": ">=5.6",
```
jaapio: Could you please change this to ^5.6 || ^7.0 current expession would also allow php v8. This doesn't exist but it will in the future.

jurruh: Fixed it!

jaapio:
Thanks, I will create a new tag later today.

jurruh:
@jaapio could you please create the tag so we can use it in our project? Thanks.

jaapio:
Just released v3.3.1

PR #128, opened 2017-11-11
Use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase

gabriel-caruso:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCases. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

I just need to bump PHPUnit version to [`^4.8.35`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-4.8.md#4835---2017-02-06), that support this namespace.

The only problem is that I could not apply this changes to the `composer.lock` file. If someone can help me, I'd be grateful :smile:


TomasVotruba:
Thanks for this!
I was about to bump dependency to v6, but there were some interdependencies that blocked it.

Actually, it doesn't make sense for such small package to have composer.lock. I'm for removing it

gabriel-caruso:
@TomasVotruba Okay. I'll rebase after you remove it, and update to `PHPUnit 6` as well.

gabriel-caruso:
@TomasVotruba I've removed, but some errors happened. Did I miss something?

TomasVotruba:
I have the same errors even without removing `composer.lock`.

There is circular dependency I try to figure out for dozen of minutes now:

- this package requires phpunit/phpunit
- phpunit/phpunit requires phpspec/prophecy
- phpspec/prophecy requires this package 

Not sure how to approach it.

TomasVotruba:
I've posted [Question to StackOverflow](https://stackoverflow.com/questions/47236948/how-to-resolve-composer-package-circular-dependency-of-external-package) to get more feedback. This is tricky one

gabriel-caruso:
I’ve faced the same loop of `requires this or that`. I’ll also do some research on that.

jaapio:
As stated in #129 I disagree wit Tomas about the removal of composer.lock.

I never researched the issue if a self referencing circulair reference. But I think it can be solved with the [replace](https://getcomposer.org/doc/04-schema.md#replace) option of composer. Need to try this myself to

TomasVotruba:
I've tried `replace`: `self.version` with no success

gabriel-caruso:
@TomasVotruba @jaapio I've rebased with master and run tests again. Still failing :sob:

TomasVotruba:
I've asked on [composer repository](https://github.com/composer/composer/issues/6805) as well,
and started [bounty on StackOverflow](https://stackoverflow.com/questions/47236948/how-to-resolve-composer-package-circular-dependency-of-external-package).

Let's :pray:

peter279k:
Perhaps it can just update the ```composer.lock``` to let this cache the correct dependencies.
I will test this at my available time.

TomasVotruba:
I've tried to remove `composer.lock` with no luck.

peter279k:
Finally, I fork the repository and it seems to be great for my Travis-ci building [log](https://travis-ci.org/peter279k/ReflectionDocBlock).
Perhaps the ```PHP 7.2``` should not be tested in ```.travias.yml``` at this time because the Travis-ci job seems to be broken.(a segmentation fault?)

TomasVotruba:
PHP 7.2 is not the issue here, **try adding PHPUnit 4.5+**: https://github.com/phpDocumentor/ReflectionDocBlock/pull/128#issuecomment-343655546

TomasVotruba:
@gabriel-caruso Could you try: https://stackoverflow.com/a/47271258/1348344

peter279k:
Ok,it seems to be worked after using the PHPUnit 4.8 and [this](https://stackoverflow.com/a/47271258/1348344) approach.
Please see more details about this job [log](https://travis-ci.org/peter279k/ReflectionDocBlock).

mleko:
Are you sure you want to test against PHP7.2? AFAIK it's not stable yet. Maybe consider adding `allow_failures`

peter279k:
I agree with @mleko, the PHP 7.2 should add in ```allow_failures```.
Because the version 7.2 is unstable at this time.

gabriel-caruso:
@TomasVotruba Ok, here is what we have right now:

- I've used [this solution](https://stackoverflow.com/a/47271258/4595097) (thanks to @mleko) to be able to update PHPUnit :white_check_mark: 
- Now that we are on [Mockery 1](https://github.com/mockery/mockery/releases/tag/1.0), I've called `Mockery::close` after each test;
- I added `doctrine/instantiator` conditional to not install the [current version, that requires `PHP 7.1`](https://github.com/doctrine/instantiator/releases/tag/1.1.0);

I think we should be ready to go.

jaapio:
However php 7.2 is not stable yet it should be possible to run the tests agains php 7.2.  I don't understand why the current build is failing for php 7.2. Looks like there is something wrong with travis at this moment?

composer.json
```diff
@@ -22,7 +22,13 @@
         "psr-4": {"phpDocumentor\\Reflection\\": ["tests/unit"]}
     },
     "require-dev": {
-        "mockery/mockery": "^0.9.4",
-        "phpunit/phpunit": "^4.4"
+        "mockery/mockery": "^1.0",
+        "phpunit/phpunit": "^6.4",
+        "doctrine/instantiator": "~1.0.5"
+    },
+        "extra": {
```
TomasVotruba: here is extra space

gabriel-caruso: Sorry, missed this one. Fixed.

jaapio:
Looks like this one is ready for merge. I don't have any comments on this one any more.

PR #131, opened 2017-11-11
Add Coding style

TomasVotruba:
This is PSR-2 based proposal for coding standard check.
All done in Travis standalone build.

Code can be fixed by:

```bash
temp/ecs/bin/ecs check src tests --fix
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .idea
 vendor
 /build
+/temp
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ install:
 script:
   - vendor/bin/phpunit --coverage-clover=coverage.clover -v
   # coding style
-  - temp/ecs/bin/ecs check src
+  - temp/ecs/bin/ecs check src tests
 
 after_script:
   - wget https://scrutinizer-ci.com/ocular.phar
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,13 @@ cache:
 
 install:
   - composer install --no-interaction
+  # coding style
+  - composer create-project symplify/easy-coding-standard temp/ecs
 
 script:
   - vendor/bin/phpunit --coverage-clover=coverage.clover -v
+  # coding style
+  - temp/ecs/bin/ecs check src
 
 after_script:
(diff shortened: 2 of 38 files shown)


TomasVotruba:
Travis is passing nicely

image

jaapio:
Thanks for this great addition. I think we should consider using an other service for our code style checks. Since it would be clearer to the users what was wrong without having to look at the build logs. 
However this is not a blocking issue for me at this moment.

TomasVotruba:
Thanks for so fast merges. It really motivating , especially this one :clap: 

Personally I use composer script shortcuts to check and fix the code. That's the best and least user demanding way to apply coding style.

I can add those, if you like.

jaapio:
I'm trying to keep up with you :-) 

Like I noticed in another PR I would like to split the CI services in to separate service. But I'm still looking for the right tools. I prefer tools that have an in repository configuration. And fail when a certain target is not reached (like travis). But I don't want to have everything in travis since those builds can be very slow. And not everyone is reading the output. 
But maybe things are improving when travis introduces build stages, which are currently in beta. 

I will try to find some time to create a project for this on github so we can share our thoughts about this in a centeralized place.

TomasVotruba:
> I'm trying to keep up with you :-)

**Much appreciation for that, it's very pleasant to work on this repository for me! :bowing_man:  **

> But I don't want to have everything in travis since those builds can be very slow.

Me neither. If we bump to PHP 7.1, I could prepare `vendor/bin/*` checks that work both on your local machine and Travis. 0-work for the end user.

> But maybe things are improving when travis introduces build stages, which are currently in beta.

I've been testing stages last 3 weeks and although it may seem *cool*, it's super slow and not very useful for open-source projects. Instead of 3 minutes, I had to wait like 10-12. I tried to be ok with that, but when you wait 12 minutes for build and it fails on missing comma, it just :angry: And I didn't want anyone contributing to feel that.

To see for yousefl, here is [stage PR](https://github.com/rectorphp/rector/pull/119) and here is [the revert](https://github.com/rectorphp/rector/commit/eed598e26c2e63cefe81382140761f8b5a31a6f7).

PR #132, opened 2017-11-13
Check if this really helps

mleko:
To be rejected.
Created just to check if I'm right
https://github.com/phpDocumentor/ReflectionDocBlock/pull/128#issuecomment-343654897


mleko:
Locally it worked
Travis reported: `Abuse detected` 😞 
Closing

PR #133, opened 2017-11-14
Allow failures on php7.2

mleko:


GrahamCampbell:
Why? If this is because of phpunit segaults, travisci have already fixed this by upgrading xdebug yesterday.

mleko:
IMHO
Since 7.2 isn't stable yet it might break again. Broken PHP shouldn't result with failed build of library. I am aware that 7.2 is RC6 marked as final.
This is purely opinion based.

jaapio:
I disagree on this. I would like to know when something is broken in php which worked before. Could be part of the feedback loop to php. Escpecially since this RC will be the last one before the final release.

PR #137, opened 2017-11-26
Bump to PHP 7.1, add typehtins

TomasVotruba:
### Why?

- majority of [PHP world big players are bumping to PHP 7.1](https://gophp71.org/)
- upcoming [PHP 7.2](https://github.com/php-ai/php-ml/pull/147) in few days
- we could add PHP 7.0/7.1 typehints via coding standards more easily in the future
- [as measured on Packagist - 67 % people](https://seld.be/notes/php-versions-stats-2017-2-edition) are on PHP 7.0+ and growing
- easier to maintain, more reliable code without testing it
- thanks to typehints, I could found and prevent some bugs already

### Related Changes

- use PHPUnit 6.0 syntax for `@expectedException`

### DIFF
--- a/src/DocBlock/Description.php
+++ b/src/DocBlock/Description.php
@@ -1,4 +1,5 @@
-<?php
+<?php declare(strict_types=1);
+
 /**
  * This file is part of phpDocumentor.
  *
@@ -62,10 +63,8 @@ class Description
      * @param string $bodyTemplate
      * @param Tag[] $tags
      */
-    public function __construct($bodyTemplate, array $tags = [])
+    public function __construct(string $bodyTemplate, array $tags = [])
     {
-        Assert::string($bodyTemplate);
-
         $this->bodyTemplate = $bodyTemplate;
         $this->tags = $tags;
     }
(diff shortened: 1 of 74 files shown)


TomasVotruba:
PR is ready to merge. Fail is caused by coveralls due to decreased constructor coverage

src/DocBlock/Description.php
```diff
@@ -83,12 +80,8 @@ public function getTags()
     /**
      * Renders this description as a string where the provided formatter will format the tags in the expected string
      * format.
-     *
-     * @param Formatter|null $formatter
-     *
-     * @return string
      */
-    public function render(Formatter $formatter = null)
+    public function render(?Formatter $formatter = null): string
```
jaapio: Is this not dubble? It already allowed null as a value.

TomasVotruba:
As for nullable, if it doesn't break the used code, it's better to use it since it creates more explicit contract.
See this short answer: https://stackoverflow.com/a/14584208/1348344

As for unused typehtins. Do you agree, it is a comment and it should add a value to the code? If so, I realize it has it's meaning pre-PHP 7.0 typehtins. But now it just old-code that does nothing. It doesn't hurt to have it (like static methods everywhere), there is just no reason to have it at all.

You can see trend in PHP from docs to typehints. In the future, there won't be comment neither for arrays (`Type[]`) and comments will really only add extra value.

jaapio:
I do agree with you that the docblocks are redundant now.
No changes are needed. Please rebase your branch so I can merge it.

TomasVotruba:
Thank you! Rebased done

Again only coveralls is failing

TomasVotruba:
:tada: :rocket:

PR #147, opened 2018-01-17
Fix return types and make it codebase more static analyser-friendly

muglug:
- Fixes a bad typehinted return type in `src/DocBlock/ExampleFinder.php`
- Fixes the same docblock as #141, plus a bunch of others
- Adds a more specific property type for `phpDocumentor\Reflection\DocBlock\Tags\BaseTag::$description` to stop PHPStan and Psalm complaining about its type within those classes.

Passing Psalm config [here](https://github.com/muglug/ReflectionDocBlock/blob/a28842cf2663e751d4f5e6e51e01fa6289fc4415/psalm.xml)

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -2,9 +2,6 @@ parameters:
     ignoreErrors:
         # false positive
         - '#Call to an undefined method object::getDocComment\(\)#'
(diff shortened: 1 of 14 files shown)


jaapio:
Thanks for your contribution. The build was not passing due a configuration error on our side. I updated this PR with #149 and added your commit there. 
So this one can be closed.

PR #150, opened 2018-01-29
Use package name in link to packagist

Nyholm:
Use the proper URLs in badges. It looks a bit sad with invalid paths: 

screen shot 2018-01-29 at 20 57 14


Nyholm:
Thank you

PR #155, opened 2018-05-17
Restore PHP version 5.5 support in 3.x branch

jordie23:
The PHP version support shouldn't have been changed on a patch version update. There's no reason to require 5.6 or 7. This breaks other projects that support 5.5 as composer won't install this package now because of this change from php 5.5 > php 5.6

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "php": "^5.6 || ^7.0",
+        "php": "^5.5 || ^5.6 || ^7.0",
         "phpdocumentor/reflection-common": "^1.0.0",
         "phpdocumentor/type-resolver": "^0.4.0",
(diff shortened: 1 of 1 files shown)


jordie23:
Not sure what that AppVeyor error is, don't think it's related to my commit.

jaapio:
We don't want to support end of life php versions. So I won't merge this one. the step forward to php 5.6 is not to big. I recommend you to upgrade your system.

jordie23:
Actually it is quite big. It requires a change from Ubuntu 14.04 to 16.04 which isn't a simple upgrade.

This is for the 3.x branch only, I'm not suggesting newer versions should support 5.5. Just this one because it started with 5.5 support. It's pretty bad to break that support arbitrarily in patch releases.

Issue #166, opened 2019-04-29
Generic tag description `0` is not being parsed as expected


filecage:
# Issue
Parsing a generic doc-block tag that has a body with contents `0` will not be interpreted as actual tag.

## Affected versions
Can reproduce in:
* 4.3.0 (latest stable)
* 5.0.0-alpha3 (latest release)
* current master (90ae750)

## Reproduction
```php
    /**
     * @arbitrary-tag 0
     */
    class Foo {
    }

    $reflection = new \ReflectionClass(\Foo::class);
    $docComment = $reflection->getDocComment();

    $factory = phpDocumentor\Reflection\DocBlockFactory::createInstance();
    $docblock = $factory->create($docComment);

    echo '>';
    echo (string) $docblock->getTags()[0];
    echo '<';
```
### Output
```
><
```

## Expected behaviour
The tag description should be treated as string with contents `0`.

## Actual behaviour
The tag description is being treated as `null`.

jaapio:
Thanks for your report. There seem to be a PR already that fixes this issue, I need to check that one :-)

filecage:
Yes, it's from one of my co-workers. :-) We've evaluated the fix in a production build already but weren't entirely sure about the behaviour change and testing setup.

It would be great if we could get this out as `4.3.1` so we can go back to stable versioning instead of using a dev version in production. We can open a PR including this fix to another branch if you wish to release to `4.x` from a different source.

jaapio:
Here you go, I cherry-picked the changes to another branch :-)

jaapio:
The check was invalid anyway, I don't expect it to break anything. And if it does it would require another fix, and test, since non are failing now

PR #167, opened 2019-04-30
Fix for generic tag description properly parsing 0 

michbeck:
Relating to #166


jaapio:
Thanks this looks good to me

michbeck:
Found that while testing our annotations with random values. My workmate opened an issue on that which i linked above. You are welcome! ;)

PR #170, opened 2019-04-30
Check on leading "[" removed as this check makes it very difficult to…

ancpru:
… implement custom tags which might want to allow or require this character in the tag body.

A leading square bracket might not be necessary for standard tags, but StandardTagFactory allows registering or overriding of tag implementations which might want to accept this character right after the tag.


mvriel:
Do you have a use-case where such behaviour is desired? Without a concrete use-case I prefer to stick to the status quo as removing it may have inadvertent side effects

jaapio:
This issue has been fixed in #200

PR #171, opened 2019-04-30
Make getter for bodyTemplate public

ancpru:
in order to make conversion of existig Description to custom Description class easier (e.g. new MyDescriptionClass($origionalDescription->getBodyTemplate(), $origionalDescription->getTags();)

### DIFF
--- a/src/DocBlock/Description.php
+++ b/src/DocBlock/Description.php
@@ -66,6 +66,15 @@ class Description
         $this->bodyTemplate = $bodyTemplate;
         $this->tags = $tags;
     }
+    
+    /**
+     * Returns the body template
+     * @return string
+     */
+    public function getBodyTemplate(): string
+    {
+        return $this->bodyTemplate;
+    }
 
     /**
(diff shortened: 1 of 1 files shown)


mvriel:
Although I would actually prefer for the `getTags()` method to be removed as the Value Object does not need to expose its data; I can see your point. Can you update the unit tests to reflect this new situation?

src/DocBlock/Description.php
```diff
@@ -66,6 +66,15 @@ public function __construct(string $bodyTemplate, array $tags = [])
         $this->bodyTemplate = $bodyTemplate;
         $this->tags = $tags;
     }
+    
+    /**
+     * Returns the body template
+     * @return string
```
mvriel: Small note: can you add a whiteline between the summary and the tags for consistency?

mvriel:
Since I understand the reasoning and I don't want to kill the PR; I have merged it and will write the test and style fix myself :)

PR #176, opened 2019-07-23
Replace dflydev/markdown suggestion

icanhazstring:
Since dflydev/markdown is abandoned, replace it with michelf/php-markdown instead.


mvriel:
Thank you for the PR!

PR #187, opened 2019-12-10
Test against php 7.4

BackEndTea:


jaapio:
Thanks for your contribution. However I'm not going to merge this, we are switching to github actions and drop usage of travis because of various issues we had with travis and a better setup we are now applying to all our repos

Issue #186, opened 2019-12-09
Generics containing spaces not parsed correctly


gronostajo:
In v4.3.2 calling `\phpDocumentor\Reflection\DocBlockFactory::create` for this phpDoc:

```
/**
 * @return array<string, Channel>
 */
```

will cause a `RuntimeException` from `\phpDocumentor\Reflection\TypeResolver::parseTypes:263` with message `'A type is missing in a collection expression'`.

The bug seems to be in `\phpDocumentor\Reflection\DocBlock\Tags\Return_::create:50`. This split:

```
$parts = preg_split('/\s+/Su', $body, 2);
```

doesn't take generics into account. `array<string,` is being read as type and `Channel>` becomes the description. Removing the space after comma solves the problem.

I've tried to determine if those spaces are actually a valid syntax and it seems like they kind-of-are:

1. [Grammar on phpdoc.org](https://docs.phpdoc.org/references/phpdoc/types.html) doesn't allow for them (on a side note, the "phpDocumentor" link in top left leads to a 404)
2. [PSR-5 grammar](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#abnf) currently doesn't support generics too
3. However the PSR-5 draft did allow for them [before Oct 16, 2018](https://github.com/phpDocumentor/fig-standards/commit/eb07c50881248dc7219de3f1ff3080362b6e8477#diff-9a10a11696dc38209c125ea9c57a565f) and spaces were allowed.
4. The file that throws exception `phpdocumentor/type-resolver/src/TypeResolver.php` does itself use spaces on lines 64-65:
    ```
    /**
     * @var array<string, string> List of recognized keywords and unto which Value Object they map
     * @psalm-var array<string, class-string<Type>>
     */
    ```

So the implementation seems to be inconsistent with grammar in PSR and usage in `phpDocumentor/TypeResolver`.

GrahamCampbell:
I think this was already addressed in v5 of this library?

kuraobi:
This broke production on our project, making Symfony serializer crash. We had to lock the version to 4.3.1 to make it work again. Maybe you should rollback 4.3.2 release since it is broken, then either fix it or just wait for v5.

jaapio:
@kuraobi I think, this was not introduced in v4.3. the tags files in v4.3 haven't been changed for years. So a more extended bug report would help instead of blaming this project without any proper feedback.

[...]

PR #191, opened 2019-12-24
Support braces in types for @return

mvriel:
In this change, I have introduced a miniature automaton-light to parse
the type from the @return body. This will prevent issues with people
using generics and other unsupported forms of types.

This change does _not_ allow for the use of Generics or similar; the
TypeResolver will still fail to resolve this type. This will remove a
breaking issue in consuming applications where a runtime exception used
to be thrown.

Please note that this change is only for @return; other tags still need
to be done. This will resolve issue #186

### DIFF
--- a/src/DocBlock/Tags/Return_.php
+++ b/src/DocBlock/Tags/Return_.php
@@ -47,10 +47,10 @@ final class Return_ extends BaseTag implements Factory\StaticMethod
         Assert::string($body);
         Assert::allNotNull([$typeResolver, $descriptionFactory]);
 
-        $parts = preg_split('/\s+/Su', $body, 2);
+        list($type, $description) = self::splitBodyIntoTypeAndTheRest($body);
 
-        $type = $typeResolver->resolve(isset($parts[0]) ? $parts[0] : '', $context);
-        $description = $descriptionFactory->create(isset($parts[1]) ? $parts[1] : '', $context);
+        $type = $typeResolver->resolve($type, $context);
+        $description = $descriptionFactory->create($description, $context);
 
         return new static($type, $description);
     }
@@ -68,5 +68,30 @@ final class Return_ extends BaseTag implements Factory\StaticMethod
     public function __toString()
     {
         return $this->type . ' ' . $this->description;
+    }
+
+    private static function splitBodyIntoTypeAndTheRest(string $body) : array
+    {
+        $type = '';
+        $nestingLevel = 0;
+        for ($i = 0; $i < strlen($body); $i++) {
+            $character = $body[$i];
+
+            if (trim($character) === '' && $nestingLevel === 0) {
+                break;
+            }
+
+            $type .= $character;
+            if (in_array($character, ['<', '(', '[', '{'])) {
+                $nestingLevel++;
+            }
+            if (in_array($character, ['>', ')', ']', '}'])) {
+                $nestingLevel--;
+            }
+        }
+
+        $description = trim(substr($body, strlen($type)));
+
+        return [$type, $description];
     }
 }
--- a/tests/unit/DocBlock/Tags/ReturnTest.php
+++ b/tests/unit/DocBlock/Tags/ReturnTest.php
@@ -112,7 +112,7 @@ class ReturnTest extends TestCase
     {
         $fixture = new Return_(new String_(), new Description('Description'));
 
-        $this->assertSame('string Description', (string)$fixture);
+        $this->assertSame('string Description', (string) $fixture);
     }
 
     /**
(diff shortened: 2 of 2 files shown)


src/DocBlock/Tags/Return_.php
```diff
@@ -69,4 +69,29 @@ public function __toString()
     {
         return $this->type . ' ' . $this->description;
     }
+
+    private static function splitBodyIntoTypeAndTheRest(string $body) : array
+    {
+        $type = '';
+        $nestingLevel = 0;
```
jaapio: We will need to support multi byte chars I think

mvriel: :thinking: Shite. Forgot :man_facepalming:

mvriel: There; amended the original test to include emoji's (which are multibyte characters)

tests/unit/DocBlock/Tags/ReturnTest.php
```diff
@@ -136,14 +137,49 @@ public function testFactoryMethod()
...
+     * @uses \phpDocumentor\Reflection\DocBlock\DescriptionFactory
+     * @uses \phpDocumentor\Reflection\TypeResolver
+     * @uses \phpDocumentor\Reflection\DocBlock\Description
+     * @uses \phpDocumentor\Reflection\Types\String_
+     * @uses \phpDocumentor\Reflection\Types\Context
+     */
+    public function testFactoryMethodWithGenericWithSpace()
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('"\array😁<string,😁 string>" is not a valid Fqsen.');
```
GrahamCampbell: Surely the 2nd emoji should be to the right of the space character, not the left?

tests/unit/DocBlock/Tags/ReturnTest.php
```diff
@@ -152,7 +188,7 @@ public function testFactoryMethodFailsIfBodyIsNotString()
 
     /**
      * @covers ::create
-     * @expectedException \InvalidArgumentException
+     * @expectedException InvalidArgumentException
```
GrahamCampbell: PHPUnit ignores imported classes when resolving this, so leaving in the leading slash is less confusing for people, otherwise they may read it as being resolved as if it were imported.

tests/unit/DocBlock/Tags/ReturnTest.php
mvriel: Added another one to the right for fun :)

tests/unit/DocBlock/Tags/ReturnTest.php
mvriel: Restored slash

tests/unit/DocBlock/Tags/ReturnTest.php
```diff
@@ -136,11 +136,46 @@ public function testFactoryMethod()
...
+     * @uses \phpDocumentor\Reflection\DocBlock\DescriptionFactory
+     * @uses \phpDocumentor\Reflection\TypeResolver
+     * @uses \phpDocumentor\Reflection\DocBlock\Description
+     * @uses \phpDocumentor\Reflection\Types\String_
+     * @uses \phpDocumentor\Reflection\Types\Context
+     */
+    public function testFactoryMethodWithGenericWithSpace()
+    {
+        $this->expectException(\InvalidArgumentException::class);
+        $this->expectExceptionMessage('"\array😁<string,😁 😁string>" is not a valid Fqsen.');
```
GrahamCampbell: We are missing a test for when we just have `'"\array<string, string>" is not a valid Fqsen.'` with a simple space there. Might be good to add that too, since it relates to the original report, and the actual usage I would expect?

src/DocBlock/Tags/Return_.php
jaapio: I would expect the usage of `mb_strlen` here. since the current string length will be incorrect when using multi byte characters. This will cause issues when you have a type only return like `@return My😁Class` I think in this case a part of the class name will be seen as description.

Also make sure that  ` trim(substr($body, strlen($type)));` when dealing with multi byte characters. 
will show you what i mean. https://3v4l.org/NBEj7

mvriel: Added extra tests to ensure Emoji's /Multibyte characters won't cause issues

tests/unit/DocBlock/Tags/ReturnTest.php
mvriel: Added extra tests to ensure this won't cause issues

mvriel:
Added extra tests to ensure Emoji's /Multibyte characters won't cause issues. I do not see anything in these tests demonstrate an issue with multibyte. At this point I think we can skip using the mb extension and merge this

PR #192, opened 2019-12-27
[4.x] Fixed Travis

GrahamCampbell:


mvriel:
Thanks Graham, I was running into the same issue and was about to look into it when I saw this PR. This is a timesaver for me!

mvriel:
Ah darn .. still not :(

https://travis-ci.org/phpDocumentor/ReflectionDocBlock/builds/630120910?utm_source=github_status&utm_medium=notification

GrahamCampbell:
It's fixed in the sense that the tests are at least running now. ;)

GrahamCampbell:
If the test suite depends on an older version of typeresolver, then we should specify it in the dev-dependencies.

PR #193, opened 2019-12-27
[4.x] Actually fixed the tests

GrahamCampbell:


mvriel:
Weird that only 7.1 fails now ..

GrahamCampbell:
Oh, I see. Sec...

GrahamCampbell:
Kinda odd we don't also have a normal run of PHP 7.1, or do prefer lowest on 7.0 (as well as normal run of 7.0). That can be handled in another PR, though.

GrahamCampbell:
Also, the code style run seems totally messed up?

mvriel:
I have merged this PR as it improves the situation; the problem seen here is an issue in the TypeResolver with multibyte character strings I think

mvriel:
Yes, @jaapio and I have been working on getting the CI pipeline in a much better shape; but we are focusing on master's of our libraries at the moment

GrahamCampbell:
Doesn't look like 5.x even does a prefer lowest run tbh. Maybe we should remove it on 4.x too?

mvriel:
Sorry, what do you mean by lowest run? I am not sure I understand

GrahamCampbell:
`--prefer-lowest`

mvriel:
@GrahamCampbell try it; let's see what happens :)

mvriel:
I am looking at the TypeResolver's multibyte handling atm

GrahamCampbell:
👍

PR #194, opened 2019-12-27
[4.x] Travis fixes

GrahamCampbell:
As you say, the good stuff is on master, but it would be useful if the builds were still at least working on 4.x. The code style run still writes a load of garbage to stderr, but at least it "works". :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,5 @@
 language: php
 
-env:
-  global:
-    - COMPOSER_FLAGS="--no-interaction"
-
 matrix:
   include:
     - php: 7.0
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,9 +1,5 @@
 language: php
 
-env:
-  global:
-    - COMPOSER_FLAGS="--no-interaction"
```
GrahamCampbell: Travis already enables this by setting an environment variable for no interaction that composer sees.

mvriel:
I will take a look at this tomorrow (or @jaapio, whoever is first ;)). I need some sleep in a bit :)

GrahamCampbell:
Finally all green. 🚀

mvriel:
Wasn't in bed yet :D Good work @GrahamCampbell ! Thanks!

GrahamCampbell:
Haha. ;)

PR #195, opened 2019-12-27
Removed bad line

GrahamCampbell:


jaapio:
Thanks again!

GrahamCampbell:
I think the failing build must be somehow related to cache? Is there a way to delete all travis' caches from the UI?

jaapio:
removed all cache and restarted the build

GrahamCampbell:
Did that restart work?

GrahamCampbell:
It doesn't seem to have re-run>

jaapio:
```
 Problem 1

    - Installation request for phpspec/prophecy 1.10.1 -> satisfiable by phpspec/prophecy[1.10.1].

    - phpspec/prophecy 1.10.1 requires phpdocumentor/reflection-docblock ^2.0|^3.0.2|^4.0|^5.0 -> no matching package found.

  Problem 2

    - phpspec/prophecy 1.10.1 requires phpdocumentor/reflection-docblock ^2.0|^3.0.2|^4.0|^5.0 -> no matching package found.

    - phpunit/phpunit 6.5.14 requires phpspec/prophecy ^1.7 -> satisfiable by phpspec/prophecy[1.10.1].

    - Installation request for phpunit/phpunit 6.5.14 -> satisfiable by phpunit/phpunit[6.5.14].
```

this is not cache, the issue is that we are a dependency of phpunit. Which is causing issues as always. This is why we were very happy with the scoped phar of phpunit v8.

GrahamCampbell:
It doesn't make sense, though. Composer resolves fine on my machine, and on the PR build.

jaapio:
there is one main difference, travis is building a commit and not a branch. So what it does is executing the clone command and after that is does:
`git checkout -qf 149120ee8fe5ab43fdb707144e56df96321591d9` 

This brings git into a detached mode, so composer does not recognize the package or any alias you are configuring. So it will fail on your machine as well when you execute that command before the composer install. :-)

GrahamCampbell:
Well... we could fix that :P ...

GrahamCampbell:
Give me a sec

jaapio:
sure, by creating a tag like `git tag 4.3.9999` ;-)

GrahamCampbell:
No, I was thinking... just get travis to actually check out the branch we are on, then hard reset to the place we need to be at.

GrahamCampbell:
Actually, I think we can just set `COMPOSER_ROOT_VERSION`.

GrahamCampbell:
I am rather hopeful this will work: https://github.com/phpDocumentor/ReflectionDocBlock/pull/196.

PR #198, opened 2020-01-03
Fix issue when processing invalid tags

jaapio:
When invalid tags are processed a null was returned causing all kind
of issues in the normal behavior of this libary. As a solution a generic tag
could be created. But that would just drop the error information in.
Therefore a new tag was introduced, `invalidTag` the tag is just like the
generic tag but does contain the error triggered during the creation of the
tag. Which might help applications like phpdocumentor to display validation issues.

### DIFF
--- a/src/DocBlock/Tags/InvalidTag.php
+++ b/src/DocBlock/Tags/InvalidTag.php
@@ -1,0 +1,72 @@
+<?php
+
+declare(strict_types=1);
+
+namespace phpDocumentor\Reflection\DocBlock\Tags;
+
+use phpDocumentor\Reflection\DocBlock\Tag;
+use Throwable;
+use Webmozart\Assert\Assert;
+
+/**
+ * This class represents an exception during the tag creation
+ *
+ * Since the internals of the library are relaying on the correct syntax of a docblock
+ * we cannot simply throw exceptions at all time because the exceptions will break the creation of a
+ * docklock. Just silently ignore the exceptions is not an option because the user as an issue to fix.
+ *
+ * This tag holds that error information until a using application is able to display it. The object wil just behave
+ * like any normal tag. So the normal application flow will not break.
+ */
+final class InvalidTag implements Tag
+{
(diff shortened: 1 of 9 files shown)


src/DocBlock/Tags/InvalidTag.php
```diff
@@ -0,0 +1,72 @@
...
+ * This class represents an exception during the tag creation
+ *
+ * Since the internals of the library are relaying on the correct syntax of a docblock
+ * we cannot simply throw exceptions at all time because the exceptions will break the creation of a
+ * docklock. Just silently ignore the exceptions is not an option because the user as an issue to fix.
+ *
+ * This tag holds that error information until a using application is able to display it. The object wil just behave
+ * like any normal tag. So the normal application flow will not break.
+ */
+final class InvalidTag implements Tag
```
mvriel: I am missing a test for this class

src/DocBlock/Tags/InvalidTag.php
```diff
@@ -0,0 +1,72 @@
...
+
+    public function getName() : string
+    {
+        return $this->name;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public static function create(string $body, string $name = '', ?Throwable $exception = null)
```
mvriel: I like how you store the exception; that way we could use this info to render an error log in phpDocumentor

mvriel: I do wonder whether it would not be better to introduce a 'withError()' method with which you can return a new version of this tag with the exception attached; this will keep the interface cleaner and will better emulate the optional character. The caller already knows it has to do something special for this tag; so for a caller it doesn't really matter.

(example: if we were to ever change the `create` definition in the interface; this would break any code that uses this exception parameter; with `withError` you avoid that problem)

jaapio:
Thanks for the review. Updated the code and merged it to master. 
We might want to merge this to v4? however, this is a breaking change?

PR #202, opened 2020-02-10
PHP 7.4 fix - Ensure the existence of the 'args' key.

drupol:
This simple fix for ensuring the existence of the 'args' key.

With PHP 7.4 on Linux and Windows, it fails without it.
Weird thing is that it doesn't break on MacOSX.

See it live here: https://github.com/ecphp/cas-bundle/actions/runs/36937260

### DIFF
--- a/src/DocBlock/Tags/InvalidTag.php
+++ b/src/DocBlock/Tags/InvalidTag.php
@@ -104,6 +104,7 @@ final class InvalidTag implements Tag
         do {
             $trace = array_map(
                 static function (array $call) use ($flatten) : array {
+                    $call += ['args' => []];
                     array_walk_recursive($call['args'], $flatten);
 
                     return $call;
--- a/src/DocBlock/Tags/InvalidTag.php
+++ b/src/DocBlock/Tags/InvalidTag.php
@@ -104,9 +104,9 @@ final class InvalidTag implements Tag
         do {
             $trace = array_map(
                 static function (array $call) use ($flatten) : array {
-                    $args = $call['args'] ?? [];
+                    $call['args'] = $call['args'] ?? [];
 
-                    array_walk_recursive($args, $flatten);
+                    array_walk_recursive($call['args'], $flatten);
 
                     return $call;
                 },
--- a/src/DocBlock/Tags/InvalidTag.php
+++ b/src/DocBlock/Tags/InvalidTag.php
@@ -104,7 +104,9 @@ final class InvalidTag implements Tag
         do {
             $trace = array_map(
                 static function (array $call) use ($flatten) : array {
-                    array_walk_recursive($call['args'] ?? [], $flatten);
+                    $args = $call['args'] ?? [];
+
+                    array_walk_recursive($args, $flatten);
 
                     return $call;
                 },
--- a/src/DocBlock/Tags/InvalidTag.php
+++ b/src/DocBlock/Tags/InvalidTag.php
@@ -104,8 +104,7 @@ final class InvalidTag implements Tag
         do {
             $trace = array_map(
                 static function (array $call) use ($flatten) : array {
-                    $call += ['args' => []];
-                    array_walk_recursive($call['args'], $flatten);
+                    array_walk_recursive($call['args'] ?? [], $flatten);
 
                     return $call;
                 },


src/DocBlock/Tags/InvalidTag.php
```diff
@@ -104,6 +104,7 @@ static function (&$value) : void {
         do {
             $trace = array_map(
                 static function (array $call) use ($flatten) : array {
+                    $call += ['args' => []];
```
GrahamCampbell: Instead of this, why not change the line below to `$call['args'] ?? []`?

GrahamCampbell:
> Weird thing is that it doesn't break on MacOSX.

Probably due to the error reporting level configuration.

src/DocBlock/Tags/InvalidTag.php
GrahamCampbell: In fact, this might be the only correct fix. I think args did exist, but was null, based on the error you linked to?

drupol: > Instead of this, why not change the line below to `$call['args'] ?? []`?

Using `$call['args'] ?? []` is also a valid solution, one or the other is fine :-)

> In fact, this might be the only correct fix. I think args did exist, but was null, based on the error you linked to?

It's ok for me then I can update.

GrahamCampbell: > Using $call['args'] ?? [] is also a valid solution, one or the other is fine :-)

Only my one is correct, I think. It also fixes the case where args is set but is null, which is the bug you were experiencing?

drupol: You are perfectly right. Thanks for pointing that out.

drupol:
Oops, it's not a valid solution.

drupol:
I have an error when using `$call['args'] ?? []`.
`PHP Fatal error:  Only variables can be passed by reference in /home/pol/dev/git/b4/cas-bundle/vendor/phpdocumentor/reflection-docblock/src/DocBlock/Tags/InvalidTag.php on line 107`

src/DocBlock/Tags/InvalidTag.php
```diff
@@ -104,7 +104,9 @@ static function (&$value) : void {
         do {
             $trace = array_map(
                 static function (array $call) use ($flatten) : array {
-                    array_walk_recursive($call['args'], $flatten);
+		    $args = $call['args'] ?? [];
```
GrahamCampbell: Please use 4 spaces for indentation, and not tabs.

drupol: It was fixed a couple of seconds after committing, sorry.

src/DocBlock/Tags/InvalidTag.php
```diff
@@ -104,7 +104,9 @@ static function (&$value) : void {
         do {
             $trace = array_map(
                 static function (array $call) use ($flatten) : array {
-                    array_walk_recursive($call['args'], $flatten);
+                    $args = $call['args'] ?? [];
```
jaapio: This is not gonna work. you are changing the original behavior here. `$args` should be set to `$call['args'] to make it work as before. better would be to use `$call['args'] = $call['args'] ?? [];` Which will overwrite `args` when it was not set. 

It is hard to test this behavior. because we are highjacking the stack trace here. Which you don't want to do in a normal case. One of the issues we were facing before this addition was that libraries could not serialize this tag.

drupol:
Investigating further... I think I have found something... 

image

jaapio:
Ha yes, that author tag is invalid, the second string between `<>` should be an e-mail address. But you still might need the suggested fix because your code should not break on the current error. This class was introduced to bypass the invalid tags ;-)

drupol:
@jaapio There is no PHPDoc issue in my sources as far as I can see, I really think in this case, we should definitely commit this fix.
I fixed the previous commit and it should be good to go now.
Let me know what you think.

jaapio:
Thanks for this fix!

drupol:
Thank you you !

vkunz:
hi,

php-7.4 introduced a new ini config flag[0] allowing to disable/enable the args index.

[0]: https://www.php.net/manual/en/migration74.other-changes.php#migration74.other-changes.ini

drupol:
> hi,
> 
> php-7.4 introduced a new ini config [flag](https://www.php.net/manual/en/migration74.other-changes.php#migration74.other-changes.ini) allowing to disable/enable the args index.

Good to know, thanks !

PR #203, opened 2020-02-10
Fix method tag for longer method names

danrot:
I have added a testcase that currently fails, although it shouldn't. There seems to be a problem with `@method` annotations, when the method name gets a little bit longer. According to Regex101 it seems to result in a [catastrophic backtrack](https://regex101.com/r/6WZ0XK/1/), which gives me the feeling that this regex is too complicated. I tried to simplify it, but couldn't find any solution that generally worked.

The problem is [this regex](https://github.com/phpDocumentor/ReflectionDocBlock/blob/1e5bd86/src/DocBlock/Tags/Method.php#L104-L133), if anybody has an idea how to improve it, I am happy to implement it.

### DIFF
--- a/src/DocBlock/Tags/Method.php
+++ b/src/DocBlock/Tags/Method.php
@@ -117,7 +117,7 @@ final class Method extends BaseTag implements Factory\StaticMethod
                             (?:[\w\|_\\\\]+)
                             # array notation
                             (?:\[\])*
-                        )*
+                        )*+
                     )
                     \s+
                 )?
--- a/tests/unit/DocBlock/Tags/MethodTest.php
+++ b/tests/unit/DocBlock/Tags/MethodTest.php
@@ -343,6 +343,37 @@ class MethodTest extends TestCase
     }
 
     /**
+     * @uses \phpDocumentor\Reflection\DocBlock\Tags\Method::<public>
+     * @uses \phpDocumentor\Reflection\TypeResolver
+     * @uses \phpDocumentor\Reflection\DocBlock\Description
+     * @uses \phpDocumentor\Reflection\Types\Context
+     *
+     * @covers ::create
+     */
+    public function testReturnTypeNoneWithLongMethodName() : void
+    {
+        $descriptionFactory = m::mock(DescriptionFactory::class);
+        $resolver           = new TypeResolver();
+        $context            = new Context('');
+
+        $description = new Description('');
+
+        $descriptionFactory->shouldReceive('create')->with('', $context)->andReturn($description);
+
+        $fixture = Method::create(
+            'myVeryLongMethodName($node)',
+            $resolver,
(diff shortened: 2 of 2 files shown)


jaapio:
Thanks for this test case, I cannot make any promises when this is fixed. But it looks like an interesting improvement first hand.

danrot:
Just FYI: This is happening when using the [Gedmo Doctrine Extensions](https://github.com/Atlantic18/DoctrineExtensions/blob/048059b/lib/Gedmo/Tree/Entity/Repository/NestedTreeRepository.php#L20), which will make in turn [Prophecy](https://github.com/phpspec/prophecy) fail when trying to mock one of these classes... So I could imagine that this error affects more people.

If anybody could give me a hint I might be able to investigate further.

Other question: If regular expressions seem to be so problematic in that regard, would it be ok to solve this with some PHP code instead?

mvriel:
> Other question: If regular expressions seem to be so problematic in that regard, would it be ok to solve this with some PHP code instead?

That would need to be thoroughly benchmarked; this component uses regex's extensive as they are very performant. This library is used in a tight loop and for some consumers thousands of times; performance is essential.

danrot:
> > Other question: If regular expressions seem to be so problematic in that regard, would it be ok to solve this with some PHP code instead?
> 
> That would need to be thoroughly benchmarked; this component uses regex's extensive as they are very performant. This library is used in a tight loop and for some consumers thousands of times; performance is essential.

But as far as I can see this call is failing because of performance issues (it seems just as if PHP stops execution of the regex, because it runs into some limit), as soon as the method name gets a little bit longer. And I am not talking about unreasonably long here, a few characters are already enough.

mvriel:
I cannot answer that question without diving into the code, for which I do not have time at this moment. I can only say, that as a design philosophy this library relies on regular expressions. 

They are generally more performant than most PHP-based solutions that we have explored in the past and changing this to a purely based PHP solution should be benchmarked to prevent a performance regression.

I am currently assuming that the issue that you encounter has to do with an error in the regex; I may be proven wrong.

danrot:
As said, I am also happy to fix this using regular expressions. However, I was not able to until now. So any tips are appreciated :slightly_smiling_face:

rvanvelzen:
Adjusting the regex like [this](https://regex101.com/r/6WZ0XK/2) should help.

danrot:
@rvanvelzen Didn't know about possessive quantifiers so far! Thank you very much, all tests are succeeding now for me :slightly_smiling_face:

mvriel:
@rvanvelzen you are a true regex wizard; I am in awe of your skills :)

mvriel:
@danrot thanks for reporting the issue and make this happen

danrot:
@mvriel Happy to give back from time to time :slightly_smiling_face:

PR #204, opened 2020-02-11
Danrot long method annotation

jaapio:
This is a POC to see what happens when we handover the return type of a method tag to our type resolver. 
First results look promising. 

refs #203


jaapio:
replaced by #343

PR #207, opened 2020-02-12
More static analysis fixes

orklah:
- Fixes small performances warning by PHP EA
- Bump a little type coverage
- remove redundant checks

### DIFF
--- a/src/DocBlock/Serializer.php
+++ b/src/DocBlock/Serializer.php
@@ -98,10 +98,7 @@ class Serializer
         return $comment . $indent . ' */';
     }
 
-    /**
-     * @return mixed
-     */
-    private function removeTrailingSpaces(string $indent, string $text)
+    private function removeTrailingSpaces(string $indent, string $text) : string
     {
         return str_replace(
             sprintf("\n%s * \n", $indent),
@@ -110,10 +107,7 @@ class Serializer
         );
     }
 
-    /**
-     * @return mixed
-     */
-    private function addAsterisksForEachLine(string $indent, string $text)
+    private function addAsterisksForEachLine(string $indent, string $text) : string
     {
         return str_replace(
             "\n",
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -8,6 +8,7 @@
     <arg value="p"/>
 
     <rule ref="phpDocumentor">
+        <exclude name="SlevomatCodingStandard.Exceptions.ReferenceThrowableOnly.ReferencedGeneralException" />
     </rule>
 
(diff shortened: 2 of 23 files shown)


orklah:
Phpstan doesn't seem to understand that $array[0] is the same value than first value of array_shift. It provoked some errors when I deleted redundant checks. I'll try to find an elegant solution later which pass both on psalm and phpstan.

I have another failure on BC Checks though. I changed the visibility of some attributes from protected to private. This shouldn't have any impact because the classes are final and this is a phpDocumentor namespace.
The only case would be a user creating its own class in phpDocumentor namespace and then trying to access a protected attribute from outside the class. This would be hideous.
However, it still fails the check. Do you want me to revert that part?

orklah:
Apparently, this was supposed to be fixed in BC Checks: https://github.com/Roave/BackwardCompatibilityCheck/issues/31

I'll investigate more

jaapio:
About the BC failures, I do agree that they are false negatives. I have to check with the maintainers why this is happening. 
Please proceed with the good work you are doing in this PR. Meanwhile I will check what is going on

orklah:
> About the BC failures, I do agree that they are false negatives. I have to check with the maintainers why this is happening.

I looked and this was supposed to be fixed on the V1 release. I don't get how this is still happening...

> Please proceed with the good work you are doing in this PR. Meanwhile I will check what is going on

Thanks :)

jaapio:
I think I found why BC checker thinks this is wrong... `\phpDocumentor\Reflection\DocBlock\Tags\See` is extending a base class, `\phpDocumentor\Reflection\DocBlock\Tags\See::$refers` could possibly be used by the parent.  
I will create an issue to ask about this specific case.

jaapio:
@orklah could you please revert the changes that break the bc check? It will take a while before the issue gets fixed in that tool. And I don't want to hold you back from doing a great job here. :-(

src/DocBlock/Serializer.php
```diff
@@ -98,10 +98,7 @@ public function getDocComment(DocBlock $docblock) : string
         return $comment . $indent . ' */';
     }
 
-    /**
-     * @return string
-     */
-    private function removeTrailingSpaces(string $indent, string $text)
+    private function removeTrailingSpaces(string $indent, string $text) : string
```
GrahamCampbell: Unrelated to this PR, but PSR-12 says that there should be no space before that colon. Maybe that code style should be applied throughout the codebase, after this PR.

jaapio: We applied a custom codestyle based on doctrine's coding standard. Because I like it this way. So no strict PSR-12 in here right now ;-)

orklah:
This should be good now :)

jaapio:
Dude, you are so fast! thanks!

PR #209, opened 2020-02-17
Fix phpdoc for DocBlockFactoryInterface::create

teohhanhui:
`DocBlockFactoryInterface::create` actually accepts an object for the `$docblock` parameter too.


jaapio:
Thanks, this looks good.

PR #212, opened 2020-02-21
change somes things for SA

orklah:
Change some things:
- Replace StaticMethod by Tag in docs
- standardize create() return type (self or ?self)
- change if(strlen($x)) to if($x === '')
and some other little fixes

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -26,5 +26,20 @@
                 <file name="src/DocBlock/StandardTagFactory.php"/>
             </errorLevel>
         </PossiblyNullArrayOffset>
+
+        <DeprecatedInterface>
+            <errorLevel type="info">
+                <!-- Will be removed in 6.0.0 issues/211 -->
+                <referencedClass name="phpDocumentor\Reflection\DocBlock\Tags\Factory\StaticMethod"/>
+            </errorLevel>
+        </DeprecatedInterface>
+
+        <RedundantConditionGivenDocblockType>
+            <errorLevel type="info">
+                <!-- Psalm manage to infer a more precise type than PHPStan. notNull assert is needed for PHPStan but
+                Psalm sees it as redundant -->
+                <directory name="src/DocBlock/Tags/"/>
+            </errorLevel>
+        </RedundantConditionGivenDocblockType>
     </issueHandlers>
 </psalm>
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -180,7 +180,7 @@ jobs:
           args: analyse src --configuration phpstan.neon
 
   psalm:
-    name: Psaml
+    name: Psalm
     runs-on: ${{ matrix.operating-system }}
     strategy:
       matrix:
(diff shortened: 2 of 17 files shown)


orklah:
https://github.com/phpDocumentor/ReflectionDocBlock/pull/212/checks?check_run_id=461109833#step:5:30
I'm pretty sure ?self is covariant with ?Tag...

Should I revert this?

jaapio:
Stricly the change from `?Tag` to `?Self` is a breaking change since you are narrowing the return type. But I do agree that isn't a break in this specific case. 

php 7.4 would even allow this native, so I think it is ok to leave it. But I'm wondering if we can make the build green again after we are merging this :-)

orklah:
I'm narrowing the declared type, but it will return the same content. Anyway, as for last time, the class is final and can't be inherited so the declared type doesn't even matter :)

I'm afraid you won't be able to have green build after this too. The obvious solution would be a new major but it's kinda overkill just for a return type.

Very big projects would maintain more than one version and work on next major while still maintaining the current versions but this is not really feasible here. I guess we're left with a "TODO list" issue with every breaking change that could be made on next major?

jaapio:
Except that this is not a bc break. So it might end up in a todo list and a bug report on the tool that does the checks. Again :blush:

GrahamCampbell:
```
 Problem parsing /github/workspace/psalm.xml:
  Error on line 33:
    Element '{https://getpsalm.org/schema/config}referencedClass': This element is not expected.
```

orklah:
> ```
>  Problem parsing /github/workspace/psalm.xml:
>   Error on line 33:
>     Element '{https://getpsalm.org/schema/config}referencedClass': This element is not expected.
> ```

Thanks :)

This Psalm version is kinda old, I'll change the ignore rule, this one doesn't recognize ignored classes

psalm.xml
```diff
@@ -30,7 +30,7 @@
         <DeprecatedInterface>
             <errorLevel type="info">
                 <!-- Will be removed in 6.0.0 issues/211 -->
-                <referencedClass name="phpDocumentor\Reflection\DocBlock\Tags\Factory\StaticMethod"/>
+                <referencedClass name="phpDocumentor/Reflection/DocBlock/Tags/Factory/StaticMethod"/>
```
GrahamCampbell: i think you need `\\`?

orklah: No, I think this was a wrong fix. I make the "windows directory separator" a lot in psalm, but this is a class name, not a path. I believe Psalm was screaming because referencedClass  was introduced in a newer version. (IIRC, the version used in CI is 3.0.X)

jaapio: Let me check, I found a way to install a newer psalm version

orklah: > Let me check, I found a way to install a newer psalm version

Nice!

jaapio:
please rebase against master, this will update psalm

orklah:
This should be good to merge. I reverted the detected BC break for now

Issue #220, opened 2020-05-18
Dealing with erroneous docblock ?


drupol:
Context: https://github.com/api-platform/core/issues/3565

**Description**  
Hello,

I just took over a new project from a client and it's based on API Platform.
When running `bin/console cache:clear -vvv`, I get this issue: https://del.dog/unupowyghe.txt

It took me a while to figure out where was the issue. I added a couple of `var_dump()` to spot the erroneous docblock.

The issue was in the following entity (_I stripped down the code to the relevant part_):

```php
<?php

namespace App\Entity;

class Foo
{
    /**
     * @var
     * @ORM\Id
     * @ORM\Column(name="FOO_ID", type="integer")
     */
    private $fooId;
}
```

Basically there is an error which is triggered because the tag `@var` without a value is an [InvalidTag](https://github.com/phpDocumentor/ReflectionDocBlock/blob/master/src/DocBlock/Tags/InvalidTag.php). As `InvalidTag` doesn't have the method `getDescription()`, it fails.

I was wondering what would be the best way to fix this for the people who, like me, are spending hours finding where such an issue could be in a huge project.

I investigated [phpDocumentor/ReflectionDocBlock](https://github.com/phpDocumentor/ReflectionDocBlock/) and changed [InvalidTag](https://github.com/phpDocumentor/ReflectionDocBlock/blob/master/src/DocBlock/Tags/InvalidTag.php) so it extends [BaseTag](https://github.com/phpDocumentor/ReflectionDocBlock/blob/master/src/DocBlock/Tags/BaseTag.php), and thus, have a [getDescription()](https://github.com/phpDocumentor/ReflectionDocBlock/blob/cd72d394ca794d3466a3b2fc09d5a6c1dc86b47e/src/DocBlock/Tags/BaseTag.php#L40) method.

Of course, when I do that, I can clear the cache of the project and I get no more issue.

I'll propose a PR very soon.

jaapio:
I think we dealt with this issue? That this needs to be fixed in symfony. Please let me know if I did overlook something here.

PR #221, opened 2020-05-18
Dealing with erroneous/incomplete docblock

drupol:
This PR:

- [x] Has origin here: https://github.com/api-platform/core/issues/3565
- [x] Provide a new test
- [x] Provide a potential fix
- [x] Does not pass the CI because of an issue in the CI. Tests are passing locally.

Related to #220


jaapio:
As discussed on slack we should not fix this here. :-)

PR #223, opened 2020-05-20
POC: Setup AST lexer & parser

jaapio:


jaapio:
Will never merge this, as I'm looking at other options to improve this library

PR #225, opened 2020-06-15
Removed unnecessary version constraint

GrahamCampbell:
Not only is it not necessary, but it got forgotten about when the PHP version was bumped.


jaapio:
thanks!

PR #226, opened 2020-06-18
Fix FQSEN resolving on see,covers,uses

jaapio:
The See, Covers and Use tags can reference also methods, properties and
constants. Which means that the FqsenResolver cannot handle those properly.
This patch fixes that issue.

To not break backwards compatibilty I had to duplicate the fix over the tags. :-(


jrfnl:
Just a little note which may or may not impact this PR: `@covers` can not use unqualified names. PHPUnit demands that these are always fully qualified.

> Please note that this annotation requires a fully-qualified class name (FQCN).

Source: https://phpunit.readthedocs.io/en/9.2/annotations.html#appendixes-annotations-covers-tables-annotations

jaapio:
I think having support for unqualified names is on the wishlist for phpunit. But they don't want to implement that themselves. According to PRS-19 these tags do have support for unqualified names.

PR #227, opened 2020-06-18
remove (from publishing) files that are not needed for end user

vladyslavstartsev:


vladyslavstartsev:
the fail in CI  is not related to the file that I've edited

jaapio:
Thanks for this cleanup

PR #233, opened 2020-06-27
Update dependabot.yml

GrahamCampbell:
Fixed indentation.


jaapio:
This file was created by dependabot, so I don't think it is invalid.

GrahamCampbell:
There is a bug in dependabot then. While the whitespace is technically wrong, a lot of YAML parsers can recover still.

PR #238, opened 2020-06-27
Delete composer.lock

GrahamCampbell:
Lock files are really only meant for apps where it matters to fix dependices. For apps, it is usual to setup composer.json to only allow versions that will not introduce new breaking changes. Moreover, your lock file is actually stopping the tests from running!


jaapio:
I do agree with you that we should remove the composer.lock, but simply remove it is not enough. We should run our tests against multiple versions of the dependencies. Same would apply for phpstan and psalm. If you have time to update this pr that would be great. But I won't merge just a remove of the composer.lock file like this.

GrahamCampbell:
> We should run our tests against multiple versions of the dependencies. 

You can do that with composer. `--prefer-lowest --prefer-stable`. Take for example Laravel's test suite.

GrahamCampbell:
This PR pretty much works perfectly, apart from that we can't test on PHP 8 because the dependencies are not ready yet.

jaapio:
I'm closing this PR, the wanted approach is described in #243. PR's for that are welcome.

PR #244, opened 2020-07-08
Fix PHP 8.0 support

GrahamCampbell:


composer.json
```diff
@@ -37,5 +37,7 @@
         "branch-alias": {
             "dev-master": "5.x-dev"
         }
-    }
+    },
+    "minimum-stability": "dev",
```
GrahamCampbell, line 41: This is needed until mockery 1.3.2 and hamcrest 2.0.1 are tagged. Note that both of these are only dev-dependencies. `webmozart/assert` 1.9.1 was tagged a few minutes ago, which was all I was waiting for, before sending this PR.

mvriel: Dev dependencies I am not worried about as that does not affect our consumers; so that is really nice :) A neat trick that can help is not to set minimum-stability but instead add `@dev` after the version constraint of your dependency that needs to be dev. And if it depends on a dev dependency you need to define those as well explicitly, but it helps prevent unexpected side effects :)

mvriel: That said; the prefer-stable helps a bit too ;)

mvriel:
Nice work Graham! Thank you.

I noticed that the PR is still WIP; that is because you want to see the CI results first? Or is there more coming? :)

composer.json
GrahamCampbell: Adding `@dev` to mockery won't be enough, because we need hamcrest too.

GrahamCampbell:
> I noticed that the PR is still WIP; that is because you want to see the CI results first?

Yep.

composer.json
GrahamCampbell: Maybe if we wait to see if hamcrest gets tagged today, we can change min stability back to stable, and add `@dev` to Mockery.

jaapio: What you could do instead is add all packages that are not tagged yet with the @dev suffix. You might need an alias for the version. https://getcomposer.org/doc/articles/aliases.md#require-inline-alias

However if it works like this. We can go with this until everything is tagged. 👌

jaapio:
Thanks this looks good!

PR #247, opened 2020-07-21
Introduce safe preg_split

jaapio:
Removes the need for extra assertions which makes the code
more readable, and reduces the overhead of an is_array check.

### DIFF
--- a/src/DocBlock/DescriptionFactory.php
+++ b/src/DocBlock/DescriptionFactory.php
@@ -14,13 +14,12 @@ declare(strict_types=1);
 namespace phpDocumentor\Reflection\DocBlock;
 
 use phpDocumentor\Reflection\Types\Context as TypeContext;
-use Webmozart\Assert\Assert;
+use phpDocumentor\Reflection\Utils;
 use function count;
 use function explode;
 use function implode;
 use function ltrim;
 use function min;
-use function preg_split;
 use function str_replace;
 use function strlen;
 use function strpos;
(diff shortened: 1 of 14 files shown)


composer.json
```diff
@@ -24,6 +24,9 @@
         "mockery/mockery": "~1.3.2"
     },
     "autoload": {
+        "files": [
+            "./src/functions.php"
```
GrahamCampbell, line 28: Having a functions file will cause issues if this package is globally installed at the same time as locally installed, and the two versions differ. This is a very real possibility with this package, since it's required (indirectly) by phpunit.

This can be avoided by replacing the function with a static class method. PHP knows how to load classes, but is terrible at autoloading functions (it has no notion of that).

GrahamCampbell: See, for example https://github.com/guzzle/guzzle/pull/2546, https://github.com/guzzle/guzzle/issues/2511.

jaapio: Thanks for highlighting this... I was aware of the fact that php is bad at loading functions. however... I think the same thing applies to classes. php will load a class just once... so if you have this library installed locally, and run phpunit globally... you will never know which classes are loaded. That's the whole reason why we cannot use PHPUnit as a composer package, appart from the conflicts. 

In the past we got the wrong files under test, once a class is loaded there is no difference for php between `\phpDocumentor\Reflection\Docblock` and `\phpDocumentor\Reflection\Docblock` globally installed. The class is just there. So if phpunit loads our `\phpDocumentor\Reflection\Docblock` from it global installation. php will never start looking for another file which includes `\phpDocumentor\Reflection\Docblock` when it is needed in a unittest. At least that was my interpretation of the issues we had. 

I will definitely move this function to a class, because you had issues in the past. :+1:

PR #250, opened 2020-08-12
Fix issue with modified backtrace

### LINKED ISSUE #249: InvalidTag::flattenExceptionBacktrace() corrupts reference arguments
If there are any arguments in the backtrace that are references, InvalidTag::flattenExceptionBacktrace() may corrupt them by converting them to strings. This came up in https://github.com/nikic/PHP-Parser/issues/687.

The underlying PHP bug/feature https://bugs.php.net/bug.php?id=79108 that allows this has been fixed/removed in PHP 8.0, but this still needs a fix in this library in the meantime. It should be possible to remove the references using something like:

    $val = $array[$x];
    unset($array[$x]);
    $array[$x] = doYourTransform($val);

jaapio:
php allows the backtrace arguments to be modified via the
backtrace. By using array_map we make sure that we do not overwrite
references.

fixes #249

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -15,4 +15,8 @@
     <rule ref="SlevomatCodingStandard.Classes.SuperfluousAbstractClassNaming.SuperfluousPrefix">
         <exclude-pattern>*/src/*/Abstract*.php</exclude-pattern>
     </rule>
+
+    <rule ref="SlevomatCodingStandard.Classes.UnusedPrivateElements.UnusedMethod">
+        <exclude-pattern>*/src/DocBlock/Tags/InvalidTag.php</exclude-pattern>
+    </rule>
 </ruleset>
--- a/src/DocBlock/Tags/InvalidTag.php
+++ b/src/DocBlock/Tags/InvalidTag.php
@@ -8,12 +8,13 @@ use Closure;
 use Exception;
 use phpDocumentor\Reflection\DocBlock\Tag;
 use ReflectionClass;
+use ReflectionException;
 use ReflectionFunction;
 use Throwable;
 use function array_map;
-use function array_walk_recursive;
 use function get_class;
 use function get_resource_type;
+use function is_array;
 use function is_object;
 use function is_resource;
 use function sprintf;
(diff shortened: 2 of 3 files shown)


phpcs.xml.dist
```diff
@@ -15,4 +15,8 @@
     <rule ref="SlevomatCodingStandard.Classes.SuperfluousAbstractClassNaming.SuperfluousPrefix">
         <exclude-pattern>*/src/*/Abstract*.php</exclude-pattern>
     </rule>
+
+    <rule ref="SlevomatCodingStandard.Classes.UnusedPrivateElements.UnusedMethod">
```
williamdes, line 19: this could be moved to he source file, but it depends on your code policy.

src/DocBlock/Tags/InvalidTag.php
```diff
@@ -80,29 +81,12 @@ private function flattenExceptionBacktrace(Throwable $exception) : void
...
-            };
-
         do {
             $trace = $exception->getTrace();
             if (isset($trace[0]['args'])) {
                 $trace = array_map(
-                    static function (array $call) use ($flatten) : array {
-                        array_walk_recursive($call['args'], $flatten);
+                    function (array $call) : array {
+                        $call['args'] = array_map([$this, 'flattenArguments'], $call['args']);
```
williamdes, line 89: Give the php version requirements I would recommend (not tested):
https://github.com/phpDocumentor/ReflectionDocBlock/blob/master/composer.json#L17
```suggestion
                        $call['args'] = array_map(function($value) { return $this->flattenArguments($value); }, $call['args']);
```

jaapio:
@williamdes thanks for the review, are you able to test this with doctum?

src/DocBlock/Tags/InvalidTag.php
GrahamCampbell: 👎 for changing this. The array-callable syntax is much shorter and clearer?

williamdes: Nope, for me this syntax is "compilation" compatible and explicit for any code searcher or analyser :)
Don't you agree?

williamdes:
> @williamdes thanks for the review, are you able to test this with doctum?

Not yet, it will take me some time to test it because I can not reproduce it locally.. 
I will need to hack a phar and try on Travis CI phpMyAdmin

phpcs.xml.dist
williamdes: My proposed change would avoid this error

src/DocBlock/Tags/InvalidTag.php
williamdes: And that would remove the need of `ref="SlevomatCodingStandard.Classes.UnusedPrivateElements.UnusedMethod">`

GrahamCampbell: Both can be formally detected. I guess it just hasn't been implemented by that code style package. psalm itself certainly recognises uses like that, in it's core stuff.

jaapio:
If you can provide a phar I can reproduce the issue in docker😁

williamdes:
> If you can provide a phar I can reproduce the issue in docker😁

Well if you succeed that would help me because I could not replicate this on my workstation or docker instances 

Follow https://github.com/nikic/PHP-Parser/issues/687#issuecomment-663074942 to replicate the issue, @nikic seems to have reproduced it

jaapio:
That's what I already did, but can you provide me a phar with the fix included?

williamdes:
> That's what I already did, but can you provide me a phar with the fix included?

Sure, here it is: https://doctum.long-term.support/releases/5.0.2-dev/doctum.phar

You can see on my manifest diff that I applied your patch https://github.com/code-lts/doctum/commit/d2695c05a607861718f06e991df4cdd2191e6335#diff-d313403d9a8ddaf7d325ee4001bd3fb6

williamdes:
The patch works fine on TravisCI ! :rocket: 
https://travis-ci.org/github/williamdes/phpmyadmintest/jobs/709799927

Issue #255, opened 2020-09-20
Inline tag handling with nested braces is not working as expected


jaapio:
Given the following test case, our Description doesn't work as expected with nested braces in inline tags. 

```
   /**
     * @uses \phpDocumentor\Reflection\DocBlock\Description
     * @uses \phpDocumentor\Reflection\DocBlock\Tags\Link
     * @uses \phpDocumentor\Reflection\DocBlock\Tags\BaseTag
     * @uses \phpDocumentor\Reflection\DocBlock\Tags\Formatter\PassthroughFormatter
     * @uses \phpDocumentor\Reflection\Types\Context
     *
     * @covers ::__construct
     * @covers ::create
     */
    public function testDescriptionCanParseStringWithInlineTagAndBraces() : void
    {
        $contents   = 'This description has a {@link http://phpdoc.org/ This contains {braces} }';
        $context    = new Context('');
        $tagFactory = m::mock(TagFactory::class);
        $tagFactory->shouldReceive('create')
            ->twice()
            ->andReturnValues(
                [
                    new LinkTag('http://phpdoc.org/', new Description('This contains {braces}')),
                ]
            );

        $factory     = new DescriptionFactory($tagFactory);
        $description = $factory->create($contents, $context);

        $this->assertSame($contents, $description->render());
        $this->assertSame('This description has a %1$s', $description->getBodyTemplate());
    }
```

tipiak75:
@jaapio do you still need any help with this issue ?

If so I might have a candidate fix to contribute. 

It involves loosening up a bit the regexp in `DescriptionFactory::lex()` so that the tag is properly matched up all the way to the last closing brace. That would allow nested braces inside the tag description without compromising any other test afaik. Would that be an acceptable solution ?

With that out of the way, I noticed another issue : the trailing whitespace between closing braces in the test's expected output is not properly rendered anymore with this fix applied, causing the test to fail. It doesn't seem wrong though, it is caused by a trim() call inside `StandardTagFactory::create()` and I am not sure that's something to be changed or fixed. 

Any thoughts ?

tipiak75:
> Even with that out of the way though, I'm still losing the trailing space between the closing braces in the provided test somewhere. Still investigating why that happens, any idea ?

[...]

PR #258, opened 2020-10-12
Candidate fix for issue #255 : loosen up inline tag splitting regexp

tipiak75:
Test output before : 
```
There was 1 failure:

1) phpDocumentor\Reflection\DocBlock\DescriptionFactoryTest::testDescriptionCanParseStringWithInlineTagAndBraces
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'This description has a {@link http://phpdoc.org/ This contains {braces} }'
+'This description has a {@link http://phpdoc.org/ This contains {braces}} }'

```
Test output after : 
```
There was 1 failure:

1) phpDocumentor\Reflection\DocBlock\DescriptionFactoryTest::testDescriptionCanParseStringWithInlineTagAndBraces
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'This description has a {@link http://phpdoc.org/ This contains {braces} }'
+'This description has a {@link http://phpdoc.org/ This contains {braces}}

```
The new test case still doesn't pass because the trailing space at the end of the inline tag got cleared in `StandardTagFactory::create()` which seems to be intended behavior (removing it solves the issue but also breaks two other tests). Discussion / guidance needed in order to move forward.

### DIFF
--- a/src/DocBlock/DescriptionFactory.php
+++ b/src/DocBlock/DescriptionFactory.php
@@ -117,7 +117,7 @@ class DescriptionFactory
                             \{
                         )
                         # Match content after the nested inline tag.
-                        [^{}]*
+                        [^{]*
                     )* # If there are more inline tags, match them as well. We use "*" since there may not be any
                        # nested inline tags.
                 )
--- a/src/DocBlock/StandardTagFactory.php
+++ b/src/DocBlock/StandardTagFactory.php
@@ -44,9 +44,10 @@ use function array_slice;
 use function call_user_func_array;
 use function count;
 use function get_class;
+use function ltrim;
 use function preg_match;
 use function strpos;
-use function ltrim;
+
 
 /**
  * Creates a Tag object given the contents of a tag.
--- a/src/DocBlock/StandardTagFactory.php
+++ b/src/DocBlock/StandardTagFactory.php
@@ -145,7 +145,7 @@ final class StandardTagFactory implements TagFactory
 
         [$tagName, $tagBody] = $this->extractTagParts($tagLine);
 
-        return $this->createTag(trim($tagBody), $tagName, $context);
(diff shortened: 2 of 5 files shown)


tipiak75:
A gentle reminder (  @jaapio ), I am still waiting for any kind of feedback as to how to tackle the subsequent problem and fix the issue here if you want me to .

Should the `trim()` call in `StandardTagFactory::create()` be run conditionnally somehow ?

jaapio:
Hi,

Thanks for your pr. I do not think we should change the behavior of this trimmed white space. People using this library are expecting this to be done.

What problem do you think it would solve to make this conditional?

tipiak75:
Hi @jaapio , thanks for your feedback. Can you elaborate on this please :

> I do not think we should change the behavior of this trimmed white space. People using this library are expecting this to be done.

My understanding is the test case currently fails because the tag rendered from the mock object : `new LinkTag('http://phpdoc.org/', new Description('This contains {braces}'))`

expectedly get rendered as : `{@link http://phpdoc.org/ This contains {braces}}`

Whereas the fixture string seems to dictate the trailing space to be rendered as final part of the link tag's description : `{@link http://phpdoc.org/ This contains {braces} }`

So, where is that trailing space supposed to come from in order for the test to pass ? 
My understanding is it was supposed to be properly captured as part of the link description, in that case the link tag description object should be instantiated with it. It might be possible, fixing the mock object output, with a small tweak to StandardTagFactory::create() alongside the use of a new, custom (non-trimming) formatter class. Would you agree ?

Perhaps it's worth mentioning in all other scenarios I've explored so far, plenty of other tests tend to break as a result of my changes.

tipiak75:
@jaapio please review this new commit (along with an extra test), this is how I think it should be handled with minimal risk of BC break.

PR #262, opened 2020-10-24
Only more tests

voku:
+ small regex optimization

---
This change is [Reviewable](https://reviewable.io/reviews/phpdocumentor/reflectiondocblock/262)


jaapio:
Thanks for these tests

PR #278, opened 2021-07-04
Include default argument value when argument has no type

sforward:
This PR fixes handling default argument values when the argument has no type specified. In that case the default value is currently omitted.

This notation was already working as expected:

`@method void myMethod(string $arg = '')`

With this PR this is also correctly handled:

`@method void myMethod($arg = '')`


jaapio:
Thanks for providing this PR, I will not merge this because we are replacing the way tags are created. This will include this feature. and allow your use case. 

You can follow #343 to see our progress.

Issue #287, opened 2021-08-01
Tests are failing on Windows / Fix line ending handling


jrfnl:
All test builds against the Windows OS are currently failing and have been for a long time.

If we look back at the last time the workflow was run completely, this can be seen in the annotations below the summary, which show that the Windows test runs all failed.
https://github.com/phpDocumentor/ReflectionDocBlock/actions/runs/370378454

![image](https://user-images.githubusercontent.com/663378/127771015-1d0e098d-9f33-4b18-ba97-f4a0823a6471.png)

This was/is "hidden" due to the `continue-on-error: true` in the "Run PHPUnit" step in the unit test job, which will still mark the job as "passed", even though in reality it _failed_.

---
<details>
  <summary><strong>Details of the test failures (click to fold out)</strong></summary>

```
There was 1 error:

1) phpDocumentor\Reflection\DocBlockFactoryTest::testTagsAreInterpretedUsingFactory
Mockery\Exception\NoMatchingExpectationException: No matching handler found for Mockery_1_phpDocumentor_Reflection_DocBlock_TagFactory::create('@author Mike van Riel <me@mikevanriel.com> This is with
  multiline description.', object(phpDocumentor\Reflection\Types\Context)). Either the method was unexpected or its arguments matched no expected argument list for this method

Objects: ( array (
  'phpDocumentor\\Reflection\\Types\\Context' => 
  array (
    'class' => 'phpDocumentor\\Reflection\\Types\\Context',
    'properties' => 
    array (
    ),
  ),
))

D:\a\ReflectionDocBlock\ReflectionDocBlock\src\DocBlockFactory.php:244
D:\a\ReflectionDocBlock\ReflectionDocBlock\src\DocBlockFactory.php:104
D:\a\ReflectionDocBlock\ReflectionDocBlock\tests\unit\DocBlockFactoryTest.php:203

--

There were 11 failures:

1) phpDocumentor\Reflection\DocBlockFactoryTest::testDescriptionsRetainFormatting
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
 phpDocumentor\Reflection\DocBlock\Description Object (
-    'bodyTemplate' => 'This is a multiline Description\r\n
-    that contains a code block.\r\n
-    \r\n
+    'bodyTemplate' => 'This is a multiline Description\n
+    that contains a code block.\n
+    \n
         See here: a CodeBlock'
     'tags' => Array ()
 )

[...]

PR #290, opened 2021-08-06
GH Actions: fail the build if a test run fails

jrfnl:
The way things were set up now in the `phpunit` job, no matter whether tests passed or failed, the workflow would always continue.

I suspect this may have been set-up this way to make sure that all variations of test runs will actually be run ?
The downside is that, while you will see a :x: for the individual build in the workflow summary, the workflow will not be marked as failed, nor will the individual test builds be marked as failed.

If we look back at the last time the workflow was run completely, this can be seen in the annotations below the summary, which show that the Windows test runs all failed.
https://github.com/phpDocumentor/ReflectionDocBlock/actions/runs/370378454

I'm proposing to change this now by:
* Removing the `continue-on-error` for the test run.
* Adding the `fail-fast` key and setting it to `false`.
    By default this key is set to `true`, which means that if any individual build within the job fails, all other builds within the job will be cancelled.
    By setting it to `false`, all builds in the matrix will still be run, but if any of them fail, the workflow will be marked as "failed".

Opening as draft to expose issue #287 properly.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -54,6 +54,7 @@ jobs:
     name: Unit tests for PHP version ${{ matrix.php-versions }} on ${{ matrix.operating-system }}
     runs-on: ${{ matrix.operating-system }}
     strategy:
+      fail-fast: false
       matrix:
         operating-system:
           - ubuntu-latest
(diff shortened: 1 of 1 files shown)


jaapio:
@jrfnl could you please update this pr?

jrfnl:
@jaapio Rebased. Still in draft due to issue #287 needing solving first.

jaapio:
@jrfnl #287 has been closed. I think this pr can be rebased and merged? Or do you want to introduce some more changes before merging this?

PR #304, opened 2021-09-27
Feature/method params

jaapio:
This pr will add variadict and reference support to the `@method` arguments. 
Since we need to expose more information now the original `getArguments` method is deprecated, and will be replaced by `getParameters` which will now return a `Param` that supports more complex notations. 

There should be no BC breaks in here.


williamdes (approved):
Looks interesting !

jaapio:
Replaced by #343

PR #315, opened 2021-12-17
Customize description class in DescriptionFactory

WinterSilence:
Added property `$this->descriptionClass` to customize description class.

### DIFF
--- a/src/DocBlock/DescriptionFactory.php
+++ b/src/DocBlock/DescriptionFactory.php
@@ -50,12 +50,16 @@ class DescriptionFactory
     /** @var TagFactory */
     private $tagFactory;
 
+    /** @var class-string */
+    private $descriptionClass;
+
     /**
      * Initializes this factory with the means to construct (inline) tags.
      */
-    public function __construct(TagFactory $tagFactory)
+    public function __construct(TagFactory $tagFactory, string $descriptionClass = Description::class)
     {
         $this->tagFactory = $tagFactory;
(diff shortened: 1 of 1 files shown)


jaapio:
I'm sorry, but I won't accept this change. As it makes the code harder to maintain. For the next time, please report an issue, and try to explain what you need. I'm happy to get any contributions, as long we can all agree on how we implement it. 

If you want to overwrite the description type you are able to extend the factory, or wrap it into an custom factory class, using the https://refactoring.guru/design-patterns/decorator/php/example 

If this is somehow related to you reported issue #316, I do recommend to have a look at some custom rendering logic. The description class exposes all the properties you need to construct a proper printed description. 

Be aware that each tag has it's own description object, which should be rendered in the same way as a method description.

WinterSilence:
@jaapio and how you planed change it? copy code from parent `create()` to change one line? (:

jaapio:
You could do it like this:

```
class myDescriptionFactory extends DescriptionFactory {
     
     public function __construct(DescriptionFactory $factory) {
          $this->factory = $factory;
     }
     
     public function create(string $contents, ?TypeContext $context = null): Description {
           $description = $this->factory->create($contents, $context);
           
           return new MyDescription($description->getBodyTemplate(), $desciption->getTags()):    
     }
}

class MyDescription extends Description
{
}

```

As you see I'm using the original factory to create a description, than I do extract all I need from description to create my custom Description class.

PR #318, opened 2021-12-18
Optimize Description::render()

WinterSilence:


jaapio:
This looks good to me, could you please explain what issue you were facing here?

WinterSilence:
@jaapio inline tags using not often - half phpdoc tags not support it

PR #319, opened 2021-12-21
Test with multiple versions of the dependencies

### LINKED ISSUE #243: [CI] Composer dependency improvements
Currently, we are building only with the versions of depencies in our `composer.lock` file while we accept a wider range version of libs in this package. After some discussion we decided that we should ajust the ci pipeline and extend the unittest matrix to check with multiple versions using something like the snippet below:

```
      - name: "Cache dependencies installed with composer"
        uses: "actions/cache@v2"
        with:
          path: "${{ steps.determine-composer-cache-directory.outputs.directory }}"
          key: "php-${{ matrix.php-version }}-composer-${{ matrix.dependencies }}-${{ hashFiles('composer.lock') }}"
          restore-keys: "php-${{ matrix.php-version }}-composer-${{ matrix.dependencies }}-"

      - name: "Install lowest dependencies from composer.json"
        if: "matrix.dependencies == 'lowest'"
        run: "composer update --no-interaction --no-progress --no-suggest --prefer-lowest"

      - name: "Install locked dependencies from composer.lock"
        if: "matrix.dependencies == 'locked'"
        run: "composer install --no-interaction --no-progress --no-suggest"

      - name: "Install highest dependencies from composer.json"
        if: "matrix.dependencies == 'highest'"
        run: "composer update --no-interaction --no-progress --no-suggest"
```
We will keep the `composer.lock` in this repo because it allows us to pin on certain versions for development. Which seems to be usefull for us at the moment. 

We need to check if other buildsteps should also use this matrix, but for now we will start with the unittests.

villfa:
Fixes #243.

The minimum version of Mockery failed with PHP 8.0 and 8.1 so I had to raise it.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -56,6 +56,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        dependencies:
+          - highest
+          - locked
+          - lowest
         operating-system:
           - ubuntu-latest
           - windows-latest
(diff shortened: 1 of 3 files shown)


jaapio:
awsome! Thanks!

PR #332, opened 2022-03-31
Support reference in method tag

villfa:
This PR makes the following docblock parseable: 

```php
/**
 * @method &foo()
 */
class Test
{
    public $values = [];
    public function &__call($name, $args)
    {
        $this->values[$name] ??= null;
        return $this->values[$name];
    }
}
```
See this sample in action: https://3v4l.org/CXuLL

The `Method` class has now a method `returnsReference(): bool`. This name comes from [ReflectionFunctionAbstract::returnsReference](https://www.php.net/manual/en/reflectionfunctionabstract.returnsreference.php).

### DIFF
--- a/src/DocBlock/Tags/Method.php
+++ b/src/DocBlock/Tags/Method.php
@@ -57,6 +57,9 @@ final class Method extends BaseTag implements Factory\StaticMethod
     /** @var Type */
     private $returnType;
 
+    /** @var bool */
+    private $returnsReference;
+
     /**
      * @param array<int, array<string, Type|string>> $arguments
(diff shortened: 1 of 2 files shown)


jaapio (approved):
Thanks this looks good.

villfa:
Thanks @jaapio for accepting this change.
Would it ok to make a new release? I plan to use it for another PR to phpDocumentor.

jaapio:
I'm ok to get an PR on phpDocumentor using a dev version of this library, if that helps you to prepare a new PR :-) 

I hope I will have some more time next week to finalize some open PR's in this repo before creating a new release.

PR #355, opened 2023-09-25
fix: render tags for summary in docblock

bshaffer:
fixes https://github.com/phpDocumentor/phpDocumentor/issues/3583

As `summary` is already a string, renders the string before passing it into the `DocBlock` class with the corresponding tags, as to maintain BC while allowing tags in the first lines of method phpdoc.


jaapio:
As discussed in the orginal issue reported on phpDocumentor, this is by design. 

So I will close this PR.

PR #357, opened 2023-10-11
Fix rendering a description when it contains escaped chars and no tags

AydinHassan:
Hi!

We are using twig templates in some doc comments (don't ask 🤣) and when lexing the `%` char is escaped. `vsprintf` unescapes normally and returns the original content correctly.

But when there are no tags an [optimisation is performed](https://github.com/phpDocumentor/ReflectionDocBlock/pull/318) which no longer calls  `vsprintf` so the description is returned with the `%` chars escaped. 

To fix I simply call `vsprintf` on the description always.

### DIFF
--- a/src/DocBlock/Description.php
+++ b/src/DocBlock/Description.php
@@ -94,7 +94,7 @@ class Description
     public function render(?Formatter $formatter = null): string
     {
         if ($this->tags === []) {
-            return $this->bodyTemplate;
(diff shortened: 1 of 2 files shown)


jaapio:
Sorry for the delay... 

Thanks for addressing this!

AydinHassan:
Nice thanks!

PR #363, opened 2024-04-08
Be more strict about type definitions on param

jaapio:
Throw on invalid type definitions and unexpected type definitions. Not all types resolved by phpstan's parser are valid for docblocks, they might in a more complex type system but I do not see how these types would ever apply to param tags.


jaapio:
@williamdes these changes should fix the reported regressions on param tags. It might be that some other regressions are in the other tags, but as we are really scratching the edge cases I think we are kinda safe to except this. 

Before the reported type was also off for both cases. Resulting in unexpected types, so I do think it's ok to cover them whenever they pop up.

williamdes:
The code looks good

> so I do think it's ok to cover them whenever they pop up.

That part I am not sure to understand

jaapio:
We do support more types, but not all tags are covered by test completely. So some supported types might not be applicable to the location they are used. The array offset is an example of this.

PR #364, opened 2024-04-09
Added test for regression on invalid types

jaapio:
Added an more extensive test suite for invalid and valid types covered by this library. This does not cover all types supported by the type-resolver but it helps to cover the most common cases.


jaapio:
@williamdes I added some more tests in this pr and a small fix that will catch issues when types are not valid. The tests are easy to extend so if you want to make sure we cover some extra situations please do create a pr. 

I think after testing this a lot, that I cannot find any regressions or breaking changes. So we should be quite good to go now.

williamdes:
Thank you for your hard work, I will update Doctum soon

PR #367, opened 2024-05-08
Bugfix: resolve issue with multiline descriptions

### LINKED ISSUE #365: Release 5.4.0 does not add our phpstan tags
We are adding phpstan tags to WordPress core stubs.
https://github.com/php-stubs/wordpress-stubs/blob/master/visitor.php

After upgrading to v5.4.0 we don't see our tags.
How to proceed?

jaapio:
The phpstan parser is not consuming the full description when parsing docblocks with a more complex description. For them it's mostlikely not an issue as phpstan doesn't use the descriptions. But it will also parse the descriptions into unexpected tags. This could be an advantage but is not according to the phpdoc spec.

Our own tokenizer is already tokenizing the docblocks into the correct parts. So all we needed to do is assume all remaining tokens in the phpstan ast belong to the description. From there our own code is able to handle this as before in v5.3.

fixes #365

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -44,6 +44,7 @@
         <NoInterfaceProperties>
             <errorLevel type="info">
                 <file name="src/DocBlock/Tags/Factory/ParamFactory.php"/>
+                <file name="src/DocBlock/Tags/Factory/AbstractPHPStanFactory.php"/>
             </errorLevel>
         </NoInterfaceProperties>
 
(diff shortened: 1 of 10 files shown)


IanDelMar:
Thank you for the rapid solution once we came up with a reproducible example! For [php-stubs/wordpress-stubs](https://github.com/php-stubs/wordpress-stubs), we still see some differences compared to v5.3.0. I'll try to pinpoint those, but I can only do so tomorrow.

PR #369, opened 2024-05-13
Hack to work around parser issue

jaapio:
Hack to see if we can get around #365


jaapio:
Didn't work as it should

PR #373, opened 2024-07-16
Add new tags

ahjdev:
This is a beta pr for checking my code.
I will add template tag for sure.

But what about psalm, phan, phpstan tags? Should I handle them too?

### DIFF
--- a/src/DocBlock/Tags/Factory/ImplementsFactory.php
+++ b/src/DocBlock/Tags/Factory/ImplementsFactory.php
@@ -5,13 +5,14 @@ declare(strict_types=1);
 namespace phpDocumentor\Reflection\DocBlock\Tags\Factory;
 
 use Webmozart\Assert\Assert;
+use phpDocumentor\Reflection\DocBlock\Tag;
 use phpDocumentor\Reflection\TypeResolver;
 use phpDocumentor\Reflection\Types\Context;
-use phpDocumentor\Reflection\DocBlock\Tag;
+use PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocTagNode;
 use phpDocumentor\Reflection\DocBlock\Tags\Implements_;
 use phpDocumentor\Reflection\DocBlock\DescriptionFactory;
-use PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocTagNode;
 use PHPStan\PhpDocParser\Ast\PhpDoc\ImplementsTagValueNode;
+use phpDocumentor\Reflection\DocBlock\Tags\TemplateImplements;
 
 /**
  * @internal This class is not part of the BC promise of this library.
@@ -32,7 +33,14 @@ class ImplementsFactory implements PHPStanFactory
         $tagValue = $node->value;
         Assert::isInstanceOf($tagValue, ImplementsTagValueNode::class);
 
-        return new Implements_(
+        $description = $tagValue->getAttribute('description');
+        if (is_string($description) === false) {
+            $description = $tagValue->description;
+        }
+
+        $class = $node->name === '@implements' ? Implements_::class : TemplateImplements::class;
+
+        return new $class(
             $this->typeResolver->createType($tagValue->type, $context),
             $this->descriptionFactory->create($tagValue->description, $context) 
         );
@@ -40,6 +48,6 @@ class ImplementsFactory implements PHPStanFactory
 
     public function supports(PhpDocTagNode $node, Context $context): bool
     {
-        return $node->value instanceof ImplementsTagValueNode && $node->name === '@implements';
+        return $node->value instanceof ImplementsTagValueNode && ($node->name === '@implements' || $node->name === '@template-implements');
     }
 }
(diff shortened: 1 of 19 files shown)


ahjdev:
I also thinks I should open a pr for type resolver repository beacuse of this classes not handled:
```
ConditionalTypeNode
ConditionalTypeForParameterNode
OffsetAccessTypeNode
ObjectShapeNode
```

jaapio:
Right now we do not focus on the prefixed tags used by psalm and phpstan. Integrations do have the option to add new tags and register a factory twice to handle tags. I don't thing we should support this out of the box. The tags are mend to use by these tools. And are not general purpose tags. 

Regarding the extra types, it would be nice to support them. But I'm wondering if we should add them to the type resolver and how. As users of the type resolver will assume that they get clear types and apart from the objectshape they are all fuzzy. Maybe we should introduce them as `PseudoType` of `Mixed`? This would allow other libraries to get the information they need but if they do not support they could interpret it as mixed?

ahjdev:
Well I add them to `PseudoType` of `Mixed` last night (In Type Resolver). As far I checked the psalm etc they have extra types e.g:
key-of<T> and value-of<T>.
I will add them too.

About prefixed tags, I was thinking at least add some important of them. (lots of them will be generics and no need to new classes)

jaapio (approved):
This looks good, if you could have a look at the phpstan issue. (Missing return in the static factory) You can just return null.

And run make code-style to see the code style issues we can merge this right away.

Thank you very much for your help

ahjdev:
> This looks good, if you could have a look at the phpstan issue. (Missing return in the static factory) You can just return null.
> 
> And run make code-style to see the code style issues we can merge this right away.
> 
> Thank you very much for your help

Ok I will fix them tommorow

ahjdev:
I guess I'm done except I did not found `(Missing return in the static factory)`.

src/DocBlock/Tags/Factory/ImplementsFactory.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        $tagValue = $node->value;
+        Assert::isInstanceOf($tagValue, ImplementsTagValueNode::class);
+
+        $description = $tagValue->getAttribute('description');
+        if (is_string($description) === false) {
+            $description = $tagValue->description;
+        }
+
+        $class = $node->name === '@implements' ? Implements_::class : TemplateImplements::class;
```
jaapio: Why would you combine these tags in 1 factory? I think it's wise to keep them separate. Maybe with a base class. A factory should always return a single type in my opinion. Otherwise this is an abstract factory

src/DocBlock/Tags/TemplateExtends.php
```diff
@@ -0,0 +1,29 @@
...
+
+namespace phpDocumentor\Reflection\DocBlock\Tags;
+
+use phpDocumentor\Reflection\Type;
+use phpDocumentor\Reflection\DocBlock\Description;
+
+/**
+ * Reflection class for a {@}template-extends tag in a Docblock.
+ */
+final class TemplateExtends extends Extends_
```
jaapio, line 22: Is Template extends a decorator on extends? And could you replace template-extends with extends? Because that's what it means when you extend a tag this way. I'm not sure that's what you intended to do.

In this project we chose often to duplicate code because tags are not the same and cannot be interchanged. If we want to share code between tags we chose to create a base class. 

I'm not aware of what template-extends does. And I couldn't find it in the phpstan documentation. So it could be that you are fully correct here.

src/DocBlock/Tags/Template.php
```diff
@@ -0,0 +1,87 @@
...
+	private $bound;
+
+	/** @var ?Type */
+	private $default;
+
+    public function __construct(string $templateName, ?Type $bound = null, ?Type $default = null, ?Description $description = null)
+    {
+        $this->name         = 'template';
+        $this->templateName = $templateName;
+        $this->bound        = $bound;
```
jaapio, line 39: What is the bound property? I don't know what it means in this context. Can we clarify that by renaming the property or add some docblock that explains the value?

jaapio (commented):
I'm sorry, I must have overlooked the things I commented on in this review. It's a bit hard to review this while I'm just on a mobile phone as I'm on holiday right now. But I didn't want to let you wait for another 3 weeks. 

It's ok if you can answer my questions. We can always fix some things after this has been merged. I will not do a release before I'm able to do some tests myself, just to be sure we do not overlook things.

src/DocBlock/Tags/Factory/ImplementsFactory.php
ahjdev: Well I just think avoid duplicate codes.
it will be easy for me to seprate them

src/DocBlock/Tags/TemplateExtends.php
ahjdev: Well read [this](https://psalm.dev/docs/annotating_code/templated_annotations)

src/DocBlock/Tags/Template.php
ahjdev: I just copied property name from phpstan but as far I know:
`template T of int`
int is bound

ahjdev:
I fix these things but why for god sake there is not a script in composer.json for phpcs-fixer

jaapio:
We are using php-code-sniffer, there is a target in the make file named "code-style" to run it. 

For historical reasons this project didn't have dev dependencies. Especially when phpunit had a direct dependency this was problematic. We could never build on a new major version without a version issue on phpunit. 
I think it would be possible now to add php-code-sniffer as a normal dev dependency. If you want to do that, please do that in a separate pr.

jaapio:
When I'm back I will merge this and add some more tests to cover this new behavior. 

Thanks for your efforts

PR #425, opened 2025-12-17
chore(deps): allow webmozart/assert v2

simPod:

PR #427, opened 2025-12-22
 chore(deps): allow webmozart/assert 2 (psalm)

Chris53897:
just to test psalm


jaapio:
as #425 is merged this one can be closed?

Issue #362, opened 2024-03-20
[DEPRECATION] Param tags without name


jaapio:
`@param` tag should always have a variable name in it's body, support for params without variable names will be dropped. 

Supported formats will be: (option 3 does not really make any sense, but will be covered)

```
@param $var your description here
@param MyType $var your description
@param $var
@param MyType $var
```

Unsupported after this deprecation has been removed:

```
@param MyType your description here
@param MyType
```

williamdes:
And reported as invalid items I guess ?

jaapio:
Yes these will be reported as `InvalidTag`

norkunas:
> support for params without type ... will be dropped.

so basically this is opposite to https://github.com/symfony/symfony/pull/54523 ?

williamdes:
> > support for params without type ... will be dropped.
> 
> so basically this is opposite to [symfony/symfony#54523](https://github.com/symfony/symfony/pull/54523) ?

the linked PR feels wrong, the types should always be present
If the param do not document anything (type hints exist), drop them
But if they do have a description, keep them in sync with native hints

Anyway, that's only my point of view 😁

jaapio:
Types are optional. I just noticed that the description of this issue is wrong. Thanks for the headsup @norkunas

kkmuffme:
@param without a variable is extremely common for documentation of actions/filters in the WP ecosystem and found not only in WP itself but probably most plugins

PR #428, opened 2025-12-22
Prepare 6.x build

### LINKED ISSUE #361: [DEPRECATION] Usage of static factory methods
Parsing types and variables is becoming more complex these days. Types are evolving faster than the limited time I have to maintain this package while a large number of projects still depends on this library to fully support all kind of type definitions. 

To keep up with the ecosystem, it was easier to switch to phpstan/parser to some of the tags. These tags are currently created by a static factory method that contains all the parsing logic for that tag. This has been replaced with a more advanced way of creating those tags using a dedicated factory that can handle the phpstan AST. 

I assume that nobody is creating tag object directly, however I cannot be sure. By deprecating the create method in v5 and removing it in v6. There is a more clear migration path for users. If you are using the tags as stand-alone object you should stop doing that and move to the normal docblock factory as shown in the examples.

jaapio:
fixes #361, #362


xabbuh:
@jaapio Is it intended to get these changes in 5.x? Asking because I wonder if in Symfony we should adapt for these changes as a bugfix on all branches supporting 5.x or if that’s for a new feature supporting 6.x.

jaapio:
Whoops, that was wrong. Sorry for that.

jaapio:
I removed all the commits from the 5.x branch. Sorry for this mistake

xabbuh:
Thank you! I’ll keep you posted if I discover any issues while allowing 6.x in the upcoming Symfony 8.1 branch.

PR #437, opened 2026-01-20
Register 'mixin' tag handler in StandardTagFactory

stof:
The `AbstractPHPStanFactory` (which is confusingly not abstract at all) has a `MixinFactory` to handle `@mixin` tags, but this was not registered in the mapping.


mspirkov:
I solved this problem here: https://github.com/phpDocumentor/ReflectionDocBlock/pull/438

PR #438, opened 2026-01-20
Fix the creation of tags with types in `StandardTagFactory`

mspirkov:
@jaapio It seems to me that you forgot to do something here: https://github.com/phpDocumentor/ReflectionDocBlock/pull/429/changes#diff-0a54b372c4e4b90d958830938d282b2fe77063cbf4a3f259b2ecf19b14f862f5L104

Together with @stof, we found this issue here: https://github.com/phpspec/prophecy/pull/660/files/c8163365fe794c184f8939fdb4b3f4c4df257270#diff-182b6b3361b404112a4866aacc525865fcf2b17c9cdcfc78d8aa7a6f75102bec
At first, I thought it was a misuse issue, but now I realize it's a problem with ReflectionDocBlock.


tests/unit/DocBlock/StandardTagFactoryTest.php
```diff
@@ -544,4 +554,36 @@ public function invalidTagProvider(): array
...
+            ['@mixin Foo', Mixin::class],
+            ['@method string do()', Method::class],
+            ['@param Foo $bar', Param::class],
+            ['@property-read Foo $bar', PropertyRead::class],
+            ['@property Foo $bar', Property::class],
+            ['@property-write Foo $bar', PropertyWrite::class],
+            ['@return string', Return_::class],
+            ['@throws Throwable', Throws::class],
+            ['@var string $var', Var_::class],
+            ['@template-covariant string', TemplateCovariant::class],
```
stof: this should probably have cases covering *all* the types registered for the phpstan tag factory in StandardTagFactory (including `@implements` for instance, which is missing here)

mspirkov: Yes, you're right. I forgot about them.

tests/unit/DocBlock/StandardTagFactoryTest.php
```diff
@@ -544,4 +557,43 @@ public function invalidTagProvider(): array
...
+            ['@property-write Foo $bar', PropertyWrite::class],
+            ['@return string', Return_::class],
+            ['@throws Throwable', Throws::class],
+            ['@var string $var', Var_::class],
+            ['@template T', Template::class],
+            ['@template-covariant T', TemplateCovariant::class],
+            ['@extends Foo<Bar>', Extends_::class],
+            ['@implements Foo<Bar>', Implements_::class],
+
+            // TODO: add factories for this tags
```
mspirkov, line 594: Maybe we should create an issue so we don't forget about it

jaapio:
Thanks to you both

tests/unit/DocBlock/StandardTagFactoryTest.php
mspirkov: https://github.com/phpDocumentor/ReflectionDocBlock/issues/439

PR #444, opened 2026-02-25
Fix parsing of tag descriptions with continuation lines starting with *

tmdk:
When a tag description contains continuation lines that begin with `*`, the `*` appears twice in the parsed output. This affects block-style `@param` descriptions (common in WordPress docblocks) where list items or `@type` alignment uses a leading star:

```php
/**
 * @param array $foo {
 *     Description of foo.
 *
 *     @type string $bar Description of bar with
 *                       * a list
 *                       * spanning *multiple* lines
 * }
 */
```

The parsed description for `$foo` would incorrectly contain `** a list` instead of `* a list`.

## Root cause

`AbstractPHPStanFactory::tokenizeLine()` worked around a PHPStan lexer limitation by splitting `TOKEN_PHPDOC_EOL` tokens that contained trailing whitespace into two tokens: a bare EOL and a separate `TOKEN_HORIZONTAL_WS`. This placed `*` in both token values. When PHPStan's `parseText()` stopped early at a tag-like token (e.g. `@type`) and rolled back, `joinUntil()` concatenated both raw token values, doubling the star.

## Fix

Prefix every continuation line with `"* "` (star + space) before tokenizing. The PHPStan lexer's `TOKEN_PHPDOC_EOL` pattern matches exactly `\n` + optional whitespace + `* ` (star + one space), so it always consumes the inserted `"* "` as part of the EOL token, leaving the original indentation as a separate subsequent token. A `trim(..., "* \t")` on every EOL token value then strips the inserted `"* "` back out, reducing each EOL to a plain `"\n"`. The manual split into a separate `TOKEN_HORIZONTAL_WS` is no longer needed.

## Changes

- `src/DocBlock/Tags/Factory/AbstractPHPStanFactory.php` — revised `tokenizeLine()`
- `tests/integration/InterpretingDocBlocksTest.php` — regression test covering the
  block-description case with embedded `@type` lines and star-prefixed continuation lines

### DIFF
--- a/src/DocBlock/Tags/Factory/AbstractPHPStanFactory.php
+++ b/src/DocBlock/Tags/Factory/AbstractPHPStanFactory.php
@@ -113,7 +113,13 @@ class AbstractPHPStanFactory implements Factory
             if ($token[Lexer::TYPE_OFFSET] === Lexer::TOKEN_PHPDOC_EOL) {
                 // Strip "* " prefix (and other horizontal whitespace) again so it doesn't and up in the
                 // description when we joinUntil() in create().
-                $token[Lexer::VALUE_OFFSET] = trim($token[Lexer::VALUE_OFFSET], "* \t");
+                $fixed[] = [
+                    Lexer::VALUE_OFFSET => trim($token[Lexer::VALUE_OFFSET], "* \t"),
+                    Lexer::TYPE_OFFSET => $token[Lexer::TYPE_OFFSET],
+                    Lexer::LINE_OFFSET => $token[Lexer::LINE_OFFSET] ?? 0,
+                ];
+
+                continue;
             }
 
             $fixed[] = $token;
--- a/src/DocBlock/Tags/Factory/AbstractPHPStanFactory.php
+++ b/src/DocBlock/Tags/Factory/AbstractPHPStanFactory.php
@@ -61,7 +61,7 @@ class AbstractPHPStanFactory implements Factory
     public function create(string $tagLine, ?TypeContext $context = null): Tag
     {
         try {
-            $tokens = $this->tokenizeLine($tagLine . "\n");
+            $tokens = $this->tokenizeLine($tagLine);
             $ast = $this->parser->parseTag($tokens);
             if (property_exists($ast->value, 'description') === true) {
                 $ast->value->setAttribute(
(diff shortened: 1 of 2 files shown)


tests/integration/InterpretingDocBlocksTest.php
```diff
@@ -489,4 +489,42 @@ public function testParamTagDescriptionIsCorrectly(): void
...
+        $docblock = $factory->create($docComment);
+
+        self::assertEquals(
+            [
+                new Param(
+                    'foo',
+                    new Array_(),
+                    false,
+                    new Description(
+                        '{' . "\n" .
```
jaapio: For readability her-doc notation would be better, but I'm not entirely sure that will work given the fact the indenting is important.

src/DocBlock/Tags/Factory/AbstractPHPStanFactory.php
```diff
@@ -104,21 +104,15 @@ public function create(string $tagLine, ?TypeContext $context = null): Tag
...
-                $fixed[] = [
-                    ltrim($token[Lexer::VALUE_OFFSET], "\n\r"),
-                    Lexer::TOKEN_HORIZONTAL_WS,
-                    ($token[2] ?? 0) + 1,
-                ];
-                continue;
+            if ($token[Lexer::TYPE_OFFSET] === Lexer::TOKEN_PHPDOC_EOL) {
+                // Strip "* " prefix (and other horizontal whitespace) again so it doesn't and up in the
+                // description when we joinUntil() in create().
+                $token[Lexer::VALUE_OFFSET] = trim($token[Lexer::VALUE_OFFSET], "* \t");
```
jaapio: However this works, I prefer not to write to `$token` as in the original code. I think that makes it more readable as there are not side effects on the token array when looping over it. I think we can simply reconstruct the token and add that to `$fixed` that's what the other code did. only your fix is different.

jaapio (changes requested):
Please have a look at the code style. 
I have a question about the implementation about how the implementation is done. But that's open for discussion.

tests/integration/InterpretingDocBlocksTest.php
tmdk: changed to heredoc

src/DocBlock/Tags/Factory/AbstractPHPStanFactory.php
tmdk: changed to modify a copy of the token instead of working directly on the token being iterated over

tmdk:
> Please have a look at the code style. I have a question about the implementation about how the implementation is done. But that's open for discussion.

I was looking for commands to run linters in the wrong places and didn't notice the Makefile! Sorry about that

jaapio:
Thanks a lot! When the pipeline passes I will merge this and create a new release

PR #446, opened 2026-03-02
PHP 8.4 fix explicit nullable deprecations

dasginganinja:
I know this change is just for an example, but this flagged as not being PHP 8.4 compatible.

Thanks!


jaapio:
Thanks, those examples are more than just a example, they are used in our tests

Issue #451, opened 2026-03-18
Fix PHP 8.4 implicit nullable deprecation warnings


danherrero:
I know there were some PHP 8.4 deprecation fixes merged in recently. I was working in a PHP 8.4 env and had some additional deprecation warnings that were flagged. I will submit a PR for the fixes after creating this issue.

PR #452, opened 2026-03-18
Fix PHP 8.4 deprecations (see phpDocumentor#451)

danherrero:
This fixes implicit nullable deprecation warnings when in an PHP 8.4 environment (see issue #451)


jaapio:
Thanks for helping out!

